PR #1, opened 2019-10-10
[DOCS] Adds a philosophy section to the guide

pzuraq:
One of the main things that makes a good modifier is if it is able
to completely close over a concept or piece of functionality, and
effectively turn a side effect into derived state. I think that's a
bit of a high level concept to teach, but it's also really useful.
I tried to capture that in this section, without going into _too_ much
detail with what we mean by "side effect" and "derived state" and all
these fancy terms and such. I also tried to make it clear that it's ok
to write modifiers that _don't_ do this, just that it's a good thing
to aim for, especially in general purpose modifiers.

In the future, this could be part of the Ember guides in a more
advanced section possibly, after the more introductory stuff about
how to write them and which basic ones are included OOTB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,148 @@ ember install ember-modifier
 ```
 
 
+Philosophy
+------------------------------------------------------------------------------
+
+Modifiers are a basic primitive for interacting with the DOM in Ember. For
+example, Ember ships with a built-in modifier, `{{on}}`:
+
+```hbs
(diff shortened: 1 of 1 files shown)


rwjblue:
Eeck, @chriskrycho I think merging this broke the lovely TOC, sorry 😭

chriskrycho:
No worries, I’ll regenerate this evening and push a PR for it!

PR #2, opened 2019-10-10
add functional modifiers

spencer516:
This brings over the functionality from `ember-functional-modifiers` with a few changes:

1. Removes the `isRemoving` flag passed to the teardown method
2. Removes Service injection functionality 
3. A teeny bit of refactoring to the Weak Map state attached to the modifier functions

### DIFF
--- a/addon/-private/functional-modifier-manager.js
+++ b/addon/-private/functional-modifier-manager.js
@@ -1,0 +1,49 @@
+import { capabilities } from '@ember/modifier';
+
+const MODIFIER_ELEMENTS = new WeakMap();
+const MODIFIER_TEARDOWNS = new WeakMap();
+
+function teardown(modifier) {
+  const teardown = MODIFIER_TEARDOWNS.get(modifier);
+
+  if (teardown && typeof teardown === 'function') {
+    teardown();
+  }
+}
+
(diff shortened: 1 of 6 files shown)


addon/-private/functional-modifier-manager.js
```diff
@@ -0,0 +1,45 @@
...
+  MODIFIER_TEARDOWNS.set(modifier, teardown);
+}
+
+export default class FunctionalModifierManager {
+  capabilities = capabilities('3.13');
+
+  createModifier(factory) {
+    const { class: fn } = factory;
+
+    return (...args) => fn(...args);
```
pzuraq: Could this return the `fn` directly, since it's not injecting anything anymore?

spencer516: I _think_ we do require this as the instance of the function is the instance used in the Weak Maps to track the element and the cleanup method.

I'll add a test that covers for this case in case someone in the future has the same thought you had!

pzuraq: ah, that totally makes sense 👍 yeah a test or even just a comment would be good

pzuraq:
Awesome, thanks a ton for the PR! 😄

Issue #5, opened 2019-10-25
Unknown blueprint: functional-modifier


AvinishSharma:
The yarn install was successful for "ember-modifier" package but when trying to generate a functional modifier, I am getting unknown blueprint error.
>$ ember g functional-modifier my-modifier
Unknown blueprint: functional-modifier
this blueprint is not avail. in the list of "Available blueprints" on my machine.

spencer516:
@AvinishSharma — that was an oversight! I just opened a PR to add that blueprint back in.

pzuraq:
Fixed and released in v1.0.2, thanks @AvinishSharma for the bug report and @spencer516 for the fix!

PR #6, opened 2019-10-28
Add blueprint

spencer516:
### Description 
This PR adds a blueprint for both Functional and Class-based modifiers. 

This largely follows the pattern of the blueprints for helpers (helper [here](https://github.com/emberjs/ember.js/tree/18067782490dc8cc812adf74de5ee9e8ad8e081e/blueprints/helper) and tests [here](https://github.com/emberjs/ember.js/tree/18067782490dc8cc812adf74de5ee9e8ad8e081e/blueprints/helper-test)) with a few changes:

- Only generates an integration test (no option for a unit test)
- Adds a flag for generating a class-based modifier and defaults to functional-based (acceptable class flags are: `-s` & `--class`)
- Does not support pre-RFC232 style tests or older versions of Mocha

### Fixes
[#5 Unknown Blueprint: functional-modifier](https://github.com/ember-modifier/ember-modifier/issues/5)

### Open Questions
Whereas a `helper` will return a value by default, is there a sensible "default" modifier behavior we could add? That way we could have a semi-meaningful test from the blueprint. 

I thought about adding an attribute or something, but that didn't feel like something you would use a modifier for. 🤔

### DIFF
--- a/blueprints/modifier-test/index.js
+++ b/blueprints/modifier-test/index.js
@@ -1,0 +1,80 @@
+'use strict';
+
+const stringUtils = require('ember-cli-string-utils');
+const isPackageMissing = require('ember-cli-is-package-missing');
+
+const useTestFrameworkDetector = require('../test-framework-detector');
+const isModuleUnificationProject = require('../module-unification').isModuleUnificationProject;
+
+const path = require('path');
+
+module.exports = useTestFrameworkDetector({
+  description: 'Generates a helper integration test or a unit test.',
+
+  fileMapTokens: function() {
+    if (isModuleUnificationProject(this.project)) {
+      return {
+        __root__(options) {
+          if (options.inRepoAddon) {
+            return path.join('packages', options.inRepoAddon, 'src');
+          }
+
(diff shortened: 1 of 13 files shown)


blueprints/modifier-test/index.js
```diff
@@ -0,0 +1,80 @@
...
+    if (isModuleUnificationProject(this.project)) {
+      return {
+        __root__(options) {
+          if (options.inRepoAddon) {
+            return path.join('packages', options.inRepoAddon, 'src');
+          }
+
+          if (options.inDummy) {
+            throw new Error("The --dummy flag isn't supported within a module unification app");
+          }
```
allthesignals: What'd you call me?

pzuraq:
Test failure appears to be because the linter is attempting to run against the blueprint files, which it shouldn't be able to do. We'll need to add an exception to them in the lint config.

spencer516:
Just pushed a change that should address eslint! (Sorry: didn't realize the tests had failed!)

PR #15, opened 2020-03-03
makeFunctionalModifier -> modifier

tomwayson:
Copying and pasting the code from functional w/ clean up example, I noticed that `makeFunctionalModifier` was undefined, and it looks like it's been renamed to just `modifier`.


pzuraq:
Thank you for the fix! Really appreciated 😄

Issue #14, opened 2020-01-27
Changelog missing


Turbo87:
There does not seem to be a `CHANGELOG.md` file in this repo and https://github.com/ember-modifier/ember-modifier/releases is also not helpful.

@pzuraq what changes were introduced in the latest releases?

Turbo87:
sorry, on second reading, this sounds more rude than intended... 😅 

I'm just trying to figure out what changed and a changelog would be very much appreciated. Happy to show you how to generate one using lerna-changelog for example :)

pzuraq:
No it's my bad, I have been meaning to learn lerna-changelog and get in the habit of using it in general for a while now and keep putting it off. I'll make it a priority in the coming weeks, now that Octane work is finally cooling off.

v1.0.3 contained a bugfix for IE11, I believe that was the only change.

elwayman02:
I can help with this, @pzuraq! I'll submit an RB that matches this:

https://github.com/elwayman02/ember-scroll-modifiers/pull/10

Then you just need `GITHUB_AUTH` and `GITHUB_TOKEN` defined locally for lerna and release-it to read, respectively, and everything works seamlessly!

Here's an example of the result:

https://github.com/elwayman02/ember-scroll-modifiers/commit/58217364699cb79a8b5c22c265db4188a2257bbf

PR #16, opened 2020-03-04
Use release-it and lerna-changelog to manage releases

elwayman02:


elwayman02:
Will address #14 once a release it cut

pzuraq:
Just a quick update, I'm planning on merging this, but looks like tests are failing due to end of life for older Node versions, so we may need to get ready for a major first. Just haven't had time to dig in

rwjblue:
Updated to use latest `npx create-rwjblue-release-it-setup` result, and rebased. Now that Node < 10 support is dropped this should be good to go.

elwayman02:
Thank you @rwjblue!

PR #9, opened 2019-11-03
WIP: TypeScript conversion

NullVoxPopuli:
Related: https://github.com/ember-modifier/ember-modifier/issues/8

This is less straightforward than I thought :(

PR #23, opened 2020-05-14
Convert the addon to TypeScript

### LINKED ISSUE #8: Missing types?
readme has information about typescript, but the whole addon is JS with no type definition :-\\

chriskrycho:
- Add `ember-cli-typescript` to build TS assets, and `expect-type` to verify the types exported publicly.

- Convert implementation and tests to TypeScript. Add type tests to constrain the public API.

- Add docstrings to the publicly exported items for better editor feedback.

- Update the README:
  - Elaborate the TS docs, including full TS examples, explanations, and important caveats.
  - Add a table of contents.
  - Fix some grammatical issues.
  - Move the explanation of what the addon *is* above its note on the original contributors.

Supersedes and therefore resolves #9, and thereby also resolves #8.


chriskrycho:
As a follow-on for this (early next week) I will add support for testing against a *range* of TypeScript versions, including `@next`, so that we can provide strong stability guarantees for consumers of the addon.

I propose the following guidance around supported TypeScript versions:

1.  Initially support TypeScript versions 3.6, 3.7, and 3.9:

    -   TypeScript 3.6 [was released](https://devblogs.microsoft.com/typescript/announcing-typescript-3-6/) August 28, 2019—just 10 days after [the release of Ember 3.12 LTS](https://blog.emberjs.com/2019/08/16/ember-3-12-released.html), so this policy would directly align the supported TypeScript versions of this project with currently supported Ember versions.
    
    -   TypeScript 3.8 is excluded because it was a fairly unstable release, and simply broke against Ember's types; as such the Typed Ember Team recommended *against* adopting it to all consumers.

2.  Maintain support for TypeScript versions aligning *roughly* to Ember's LTS cycle. TypeScript releases occur quarterly, which means every other TypeScript release happens at approximately the same time as an Ember LTS. This **does not** commit the addon to supporting any *particular* TypeScript version: if the version has serious stability issues, especially related to Ember's types (as did 2.9, 3.1, and 3.8), we may decline to support it.

3.  Define the requirements for supporting *new* TypeScript versions as follows: support a new version—

    -   *does not require a release* if there are no required changes; but the documented range of supported versions should be updated
    -   *requires a bugfix release* if the types could be changed in a backwards-compatible way (that is: updating the type definitions to support the new version continues to pass the type tests for the supported range)
    -   *requires a breaking change* if the types cannot be changed in a backwards-compatible way

4.  Given (2) and (3), commit not to making breaking changes for *only* TypeScript, to minimize churn and impact on the ecosystem. Instead, align TypeScript-related breaking changes to when an Ember LTS goes out of support.

cc. @pzuraq @dfreeman @jamescdavis @rwjblue

rwjblue:
In general, that policy seems good to me overall though I don't know that the 4th point will work in practice (if users can't make a modifier using ember-cli-typescript and we can fix it, we _probably_ should do the major bump release regardless of Ember's LTS cycle).

addon/-private/class/modifier-manager.ts
```diff
@@ -0,0 +1,70 @@
...
+      return;
+    }
+
+    let meta = Ember.meta(instance);
+
+    meta.setSourceDestroying();
+    instance[DESTROYING] = true;
+
+    schedule("actions", instance, instance.willDestroy);
+    schedule("destroy", undefined, scheduleDestroy, instance, meta);
```
rwjblue, line 55: Can you create an issue to remove all of this code in favor of using the destroyables RFC (emberjs/rfcs#580) by way of https://github.com/ember-polyfills/ember-destroyable-polyfill?

addon/-private/class/modifier.ts
```diff
@@ -0,0 +1,126 @@
...
+   */
+  willDestroy() {}
+
+  get isDestroying() {
+    return this[DESTROYING];
+  }
+
+  get isDestroyed() {
+    return this[DESTROYED];
+  }
```
rwjblue, line 99: Similar to my comment before, I'd like to use `isDestroying` / `isDestroyed` methods from `@ember/destroyable` here instead of having to do it ourselves with the fake symbol setup.

addon/-private/class/modifier-manager.ts
```diff
@@ -0,0 +1,70 @@
...
+import { ModifierArgs } from "ember-modifier/-private/interfaces";
+
+class ClassBasedModifierManager {
+  capabilities = capabilities("3.13");
+
+  createModifier(
+    factory: { owner: unknown; class: typeof ClassBasedModifier },
+    args: ModifierArgs
+  ) {
+    let { owner, class: modifier } = factory;
```
rwjblue, line 20: This is kinda wild, we should not be getting the `owner` from this argument 🤔 

See the modifier manager API RFC https://github.com/emberjs/rfcs/blob/master/text/0373-Element-Modifier-Managers.md#createmodifier, which says:

> The first argument passed to createModifier is the result returned from the factoryFor API. It contains a class property, which gives you the the raw class (the default export from app/modifiers/foo.js) and a create function that can be used to instantiate the class with any registered injections, merging them with any additional properties that are passed.

---

Since this is a carry over from existing implementation I think an inline comment with a link to an issue is fine for now. Ember should also be updated to _not_ pass the `owner` like this (though that will have to go through a deprecation cycle).

cc @pzuraq

addon/-private/class/modifier.ts
```diff
@@ -0,0 +1,126 @@
...
+ *     doSomethingWithElement(this.element); // TYPE ERROR!
+ *   }
+ * }
+ * ```
+ */
+export type InTeardown<M extends ClassBasedModifier> = Omit<M, "element"> & {
+  element: null;
+};
+
+setModifierManager(() => Manager, ClassBasedModifier);
```
rwjblue, line 126: Any reason not to do this all in the export?

```ts
export default setModifierManager(() => Manager, class Modifier {

});
```

chriskrycho:
@rwjblue that's fair for sure; that was *definitely* the most conservative aspect of this. (To be clear the only thing being blocked would be *upgrading to a new TS version*, rather than continuing to use e-c-ts as normal, but even so!)

On reflection, though, maybe we aim to be that conservative only around *dropping* supported versions? It's not so big a deal to cut a breaking change that has a minor change; I take it to be a *slightly* (?) bigger deal to remove a supported version from the matrix.

addon/-private/functional/modifier.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * The modifier can also optionally return a *destructor*. The destructor
+ * function will be run just before the next update, and when the element is
+ * being removed entirely. It should generally clean up the changes that the
+ * modifier made in the first place.
+ *
+ * @param fn The function which defines the modifier.
+ */
+export default function modifier(fn: FunctionalModifier): unknown {
+  return setModifierManager(managerFor, fn);
```
rwjblue, line 38: The functional modifier manager does not use the owner _at all_ (the original logic for `managerFor` was created to support passing services to functional modifiers, but that functionality was removed while merging into ember-modifier), this should be simplified to:

```
return setModifierManager(() => Manager, fn);
```

The functional modifier manager file should be updated to do:

```ts
export const Manager = new FunctionalManager();
```

---

This is an important but non-blocking issue, I think making a new issue and linking with an inline comment is fine for now.

cc @pzuraq

addon/-private/utils/symbol.ts
```diff
@@ -11,7 +11,7 @@
  * https://github.com/emberjs/data/blob/master/packages/store/addon/-private/ts-interfaces/utils/symbol.ts
  *
  */
-export const symbol =
-  typeof Symbol !== 'undefined'
+export const symbol: (key: string) => symbol | string =
```
rwjblue, line 14: When we migrate to destroyables, this whole file can be deleted

addon/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default, InTeardown } from "./-private/class/modifier";
```
rwjblue, line 1: Is `InTeardown` externally visible? If so, how?

package.json
```diff
@@ -18,17 +18,25 @@
     "lint:js": "eslint .",
     "start": "ember serve",
     "test": "ember test",
-    "test:all": "ember try:each"
+    "test:all": "ember try:each",
+    "prepublishOnly": "ember ts:precompile",
+    "postpublish": "ember ts:clean"
   },
   "dependencies": {
     "ember-cli-babel": "^7.11.1",
```
rwjblue, line 26: Please update this to latest (to prep for a future ember-cli-typescript@4 update).

package.json
```diff
@@ -18,17 +18,25 @@
...
     "ember-cli-normalize-entity-name": "^1.0.0",
     "ember-cli-string-utils": "^1.1.0",
+    "ember-cli-typescript": "^3.1.3",
     "ember-modifier-manager-polyfill": "^1.2.0"
   },
   "devDependencies": {
     "@ember/optional-features": "^1.0.0",
+    "@types/ember": "^3.1.2",
+    "@types/ember-qunit": "^3.4.8",
+    "@types/ember__test-helpers": "^0.7.10",
```
rwjblue, line 37: `__😭__`

tsconfig.json
```diff
@@ -0,0 +1,56 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
+    "allowJs": true,
```
rwjblue, line 4: Why do we need `allowJs`? I've had issues with TS errors being ignored until publish time (and breaking publishing automation) because `allowJs` is explicitly set to `false` by `ts:precompile` here

https://github.com/typed-ember/ember-cli-typescript/blob/v3.1.3/ts/lib/commands/precompile.ts#L31

types/ember-private/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  export namespace Ember {
+    export interface Meta {
+      setSourceDestroying(): void;
+      setSourceDestroyed(): void;
+    }
+
+    export function meta(obj: object): Meta;
+    export function destroy(obj: object): void;
+  }
+}
```
rwjblue, line 17: This can be deleted when we move to using destroyables

types/global.d.ts
```diff
@@ -0,0 +1,6 @@
+// Types for compiled templates
+declare module 'ember-modifier/templates/*' {
```
rwjblue, line 2: This confuses me, can you explain what this does? Also, AFAIK this package doesn't have a `addon/templates` folder...

package.json
chriskrycho: THE BEST

tsconfig.json
chriskrycho: This is a good catch. I will have to dig into the history to see why we set that default, and likely change it – upcoming 4.0 release would be a good time.

types/global.d.ts
chriskrycho: Great catch; I'll remove it from this PR. In general, this is generated by e-c-ts so that folks don't get type errors for importing templates, e.g. to set `layout` on components in addons.

rwjblue:
To be clear, I think we should work on landing this without the refactors that I mention in my comments (though we should still make a todo issue and link to it in the code), but there are some more fundamental things that I do think we should fixup before landing (allowJs and ember-cli-babel version)

addon/index.ts
chriskrycho: It is, exported as a type utility so that users can do this to work around the limitations of the way the current API interacts with the type system if so desired:

```ts
import Modifier, { InTeardown } from 'ember-modifier';

export default class UsesWillDestroy {
  willDestroy(this: InTeardown<UsesWillDestroy>) {
    this.element // null
  }
}
```

rwjblue:
> On reflection, though, maybe we aim to be that conservative only around dropping supported versions? It's not so big a deal to cut a breaking change that has a minor change; I take it to be a slightly (?) bigger deal to remove a supported version from the matrix.

Ya, this makes the most sense to me. Revving majors is "fine" (for the most part), but we do have to be somewhat careful about the poor merging behavior that ember-cli gives us by default (though that is an issue for a different repo 😝).

README.md
```diff
@@ -567,7 +635,160 @@ export default class ScrollPositionModifier extends Modifier<ScrollPositionModif
 }
 ```
 
-See [this pull request comment](https://github.com/sukima/ember-class-based-modifier/pull/5#discussion_r326687943) for a full discussion about using TypeScript with your Modifiers.
+However, while doing so is slightly more convenient, it means you get *much worse* feedback in tests or at runtime if someone passes the wrong kind of arguments to your modifier.
```
NullVoxPopuli, line 638: This is great documentation! <3
Pointing out all the caveats is great! -- like, not assuming people already know the caveats (like, I often do (assume people know things))

addon/index.ts
rwjblue: Ummm, yuck 😝 

Seriously though, we should strongly consider removing the nulling of `this.element`. I think in practical terms it is not doing us any favors and it is extremely unlikely to be a reasonable stop gap for memory leaks (which is likely the reason it was added).

README.md
```diff
@@ -567,7 +635,160 @@ export default class ScrollPositionModifier extends Modifier<ScrollPositionModif
...
+
+```ts
+import Modifier, { InTeardown } from 'ember-modifier';
+
+function doSomethingWithElement(el: Element) {
+  // ...
+}
+
+export default class CorrectlyTypedModifier extends Modifier {
+  willRemove(this: InTeardown<CorrectlyTypedModifier>) {
```
NullVoxPopuli, line 654: what are the odds the TS team lets this be inherited from the super class?

chriskrycho: Heh, almost certainly not—it'd be pretty weird, and likely broken, in part because you'd end up throwing away data about the subclass… and since a `Parent` almost always has fewer properties, methods, etc. than does `Subclass`, that'd be a pretty terrible outcome. (I actually experimented with that the other day, realized it didn't work and then had to think through *why* it didn't.)

addon/-private/class/modifier-manager.ts
chriskrycho: Definitely!

addon/index.ts
chriskrycho: I strongly agree, and will be opening an issue with proposed changes to the API *after* we land this in a stable release. (🙅 to refactors that also introduce changes, as I know you feel as well.)

addon/-private/class/modifier.ts
chriskrycho: Seems reasonable to me. I'll flag it up for one of the follow-up cleanups.

package.json
chriskrycho: Done!

tsconfig.json
chriskrycho: Fixed!

addon/-private/class/modifier-manager.ts
chriskrycho: Tracking conversion to using the polyfill in #24.

addon/-private/class/modifier-manager.ts
chriskrycho: Tracking at #25

addon/-private/functional/modifier.ts
chriskrycho: Tracking at #26

addon/-private/class/modifier.ts
chriskrycho: Actually, on further consideration: yeah, we do actually have a good reason to define it independently: consumers need to be able to import the class to subclass!

rwjblue: Hmm, that doesn't get changed in my example. `setModifierManager` returns the second arg, so `setModifierManager(someFunc, Foo)` results in `Foo` (and I think the types do show that, but I could be wrong).

chriskrycho: Interesting; I misunderstood the RFC—I thought (and the types I did for this repo, since there are none public at present) that it returned `void`. I’ll update the local type definitions to match that, so that when we upstream they’re correct.

The other consideration, though, is the `InTeardown` utility/hack type. Since there would be no named type for it to refer to for defining itself, there would be no way to make it work. Given that, I think I’d suggest we keep it as is for now, and then switch it to that form if/when we make a breaking release with some of the other changes we’ve gathered here?

rwjblue: Yep, definitely seems good

.gitignore
```diff
@@ -23,3 +23,7 @@
 /.node_modules.ember-try/
 /bower.json.ember-try
 /package.json.ember-try
+
+# lockfiles -- this is a library; lock files are more apt to mislead than help
```
NullVoxPopuli, line 27: <3
This is the same approach ruby/gems have!

rwjblue: Hmm, ya don’t think we should do this. All of our ember-try scenarios already test with floating deps / no lock file, and our linting / basic tests should always be running against known good deps.

chriskrycho: @rwjblue, I'm happy to invert it; this was just originally the recommendation I saw for libraries from Yarn itself. Will push a version with the lockfile!

rwjblue: > this was just originally the recommendation I saw for libraries from Yarn itself

Ya, I also recall some confusion around that at initial release time, but I believe it is pretty well cleared up at this point. Let’s keep this close to the ember-cli addon defaults (not sure if the stock one is good enough for us, but we can make fewer tweaks).

Some references:

https://classic.yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control


https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

chriskrycho: Resolved as (force pushed) 8e3bcd2!

addon/index.ts
chriskrycho: Tracking *this* follow-up at #29.

types/dummy/index.d.ts
```diff
@@ -0,0 +1 @@
+
```
rwjblue, line 1: Do we need this empty file? What does it do for us?

chriskrycho: We don't! I'll remove it!

chriskrycho:
@rwjblue my thought on landing this was that we would do basically this sequence:

1. Land this PR.
2. Land the PR I'm working on right now which supports testing types against the support matrix of stable versions.
3. Define the policy issues—probably in parallel with (2).
4. Publish.
5. Refactor.

That said, I'm perfectly fine with switching the order and doing (5) before (3), and accordingly _not_ publishing types.

rwjblue:
I don’t have a strong preference other than that we have to deal with the SemVer / TS issue before publishing types officially. I believe that we are on the same page here though, so I’m perfectly happy to go in whatever order you prefer.

chriskrycho:
100% agreed that we have to nail that down before publishing! See #30 – I'll publish a draft there of the revised proposal I came up with after conversation here and talking with other Typed Ember folks.

PR #30, opened 2020-05-19
Support linting TS, and test types

### LINKED ISSUE #27: Add TS linting with @typescript-eslint
Switching to TS means the current linting setup now ignores every file in the repository. Add @typescript-eslint and give this a reasonable default configuration.

(Follow-on issue from #23.)

### LINKED ISSUE #28: Specify and test supported TS versions
For the sake of ecosystem stability, it is important that we treat TypeScript versions carefully.

- formally define it (see [discussion in #23](https://github.com/ember-modifier/ember-modifier/pull/23#issuecomment-628901297) for baseline)
- add CI infrastructure to enforce it

(Follow-on from #23)

chriskrycho:
Lint with `@typescript-eslint`, and fix lint errors

- Use `@typescript-eslint` config for TS code (resolves #27)
- Add a TS types config for ember-try (resolves #28)
- Fix all lint errors flagged by adding linting.
- Constrain and warn about faking `unique symbol`
- Add test scenarios for (proposed) supported TS versions

---

Original writeup:

1. Does this config seem reasonable? Also, any suggestions for improvement?
2. How exactly should we integrate this `ember-try` config into CI?

@rwjblue @pzuraq @dfreeman @jamescdavis


chriskrycho:
I've developed (in conversation with @dfreeman and @jamescdavis) a revised proposal on types stability, which I'm happy to include here (i.e. in the README) or in a separate PR. I suspect this one will be generally agreeable to everyone!

.eslintrc.js
```diff
@@ -1,22 +1,26 @@
 module.exports = {
   root: true,
-  parser: 'babel-eslint',
+  parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module'
   },
```
rwjblue, line 7: This ensures editor plugins can run from any CWD and find the right tsconfig.json:

```suggestion
  parserOptions: {
    ecmaVersion: 2018,
    sourceType: 'module',
    project: './tsconfig.json',
    // allows eslint from any dir
    tsconfigRootDir: __dirname,
    sourceType: 'module',
  },
```

.eslintrc.js
```diff
@@ -47,7 +51,36 @@ module.exports = {
       plugins: ['node'],
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
```
rwjblue, line 52:
```suggestion
      extends: ['plugin:node/recommended'],
      rules: {
```

.eslintrc.js
```diff
@@ -47,7 +51,36 @@ module.exports = {
...
+        "@typescript-eslint/no-var-requires": ["off"],
+        "no-dupe-else-if": ["off"],
+        "no-import-assign": ["off"],
+        "no-setter-return": ["off"],
+        "@typescript-eslint/camelcase": ["off"]
       })
-    }
+    },
+    {
+      files: ["*.ts"],
```
rwjblue, line 63: Seems like this should be 

```suggestion
      files: ["**/*.ts"],
```

.eslintrc.js
```diff
@@ -47,7 +51,36 @@ module.exports = {
...
+        "@typescript-eslint/explicit-function-return-type": [
+          "error",
+          {
+            allowExpressions: true,
+            allowHigherOrderFunctions: true
+          }
+        ],
+        "no-dupe-else-if": ["off"],
+        "no-import-assign": ["off"],
+        "no-setter-return": ["off"],
```
rwjblue, line 74: Seems like these are changed in all scenarios, can you remove them from overrides and hoist to the top level `rules`?

.eslintrc.js
```diff
@@ -47,7 +51,36 @@ module.exports = {
...
+      }
+    },
+    {
+      files: ["tests/dummy/app/router.js"],
```
rwjblue, line 78: Why do we need this?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
```
rwjblue, line 3: In general, I don't think we should commit editor specific files to the repo. Is using the local TS not the default behavior?

config/ember-try-typescript.js
```diff
@@ -0,0 +1,33 @@
+module.exports = () => ({
```
rwjblue, line 1: FWIW, if you don't need dynamism ember-try still supports returning a pojo (e.g. no need for this arrow function).

.eslintrc.js
```diff
@@ -1,22 +1,26 @@
...
+    '@typescript-eslint',
     'ember'
   ],
   extends: [
     'eslint:recommended',
-    'plugin:ember/recommended'
+    'plugin:@typescript-eslint/eslint-recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:ember/recommended',
+    'prettier/@typescript-eslint',
```
rwjblue, line 17: `prettier/@typescript-eslint` is brought in here, but prettier itself is not being ran. I think we either need to remove this (and do prettier separately) or add `plugin:prettier/recommended` to actually enforce prettier.

package.json
```diff
@@ -58,6 +60,8 @@
     "ember-source": "~3.13.0",
     "ember-source-channel-url": "^2.0.1",
     "ember-try": "^1.2.1",
+    "eslint": "^7.0.0",
+    "eslint-config-prettier": "^6.11.0",
```
rwjblue, line 64: See other comments, but we either need to bring in the other cadre of prettier packages (or read through the details over in https://github.com/emberjs/rfcs/pull/628 😉) in this PR (`eslint-plugin-prettier` + `prettier`).

tests/dummy/app/router.js
```diff
@@ -6,6 +6,7 @@ const Router = EmberRouter.extend({
   rootURL: config.rootURL
 });
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
 Router.map(function() {
 });
```
rwjblue, line 11: FWIW, I don't know that we actually need to have this. Probably can just delete it?

.vscode/settings.json
chriskrycho: I agree, and sadly it is *not*. (Which boggles my mind.)

config/ember-try-typescript.js
chriskrycho: Ah, perfect; I'll change it to that!

rwjblue:
> How exactly should we integrate this ember-try config into CI?

How long does it take to run them all? Seems like it should probably be fairly quick (vs doing a full `ember test` in each). I'd probably start with adding a single CI job that effectively does `ember try:each --config ./path/to-try-config-file.js`.

chriskrycho:
That was roughly what I was leaning toward as well. It's very quick for a library this size; the main (nearly the entire) time cost is installing the TS versions (as you'd expect).

.eslintrc.js
chriskrycho: Ah, yes. 👍 I didn't realize it ended up hitting them all. I need to figure out _why_ these are an issue as well. 🤔

.eslintrc.js
chriskrycho: On reflection, we actually *can* make it `router.ts` (though we could also just delete it for this scenario I think)—we don't in general touch it because the Ember generator doesn't work with the router b/c of how it does the transform.

rwjblue: Sure but I more meant that I don't think we need this override at all (I think once you hoist the `no-dupe-else-if`, `no-import-assign`, `no-setter-return` up to the top `rules` you won't need this at all).

.vscode/settings.json
rwjblue: Uh, wut. How does this make sense? Have we searched for a bug report? This seems fundamentally invalid to me... 🤔

.eslintrc.js
chriskrycho: Oh, yes, agreed. Meant to say as much but was typing and playing D&D simultaneously. 😂

.vscode/settings.json
chriskrycho: You and me and everyone else. It's a standing WHUT among VS Code TS users.

rwjblue:
FWIW, CI is getting hosed due to Node 8

```
error semver@7.3.2: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command "yarn install --non-interactive" failed and exited with 1 during .
```

Probably time to drop it (in a separate PR), then prep this batch of TS changes for the next major...

chriskrycho:
Ah, that makes sense. Will do.

.eslintrc.js
chriskrycho: I'll remove prettier in *this* one and add it separately.

.eslintrc.js
chriskrycho: This appears to have been doing this because [using `extends` in `overrides` is unsupported](https://github.com/eslint/eslint/issues/8813).

.eslintrc.js
chriskrycho: When I used this we end up with:

```
/Users/ckrycho/dev/ember-modifier/ember-modifier/index.js
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: index.js.
The file must be included in at least one of the projects provided
```

.eslintrc.js
```diff
@@ -1,22 +1,32 @@
...
-  parser: 'babel-eslint',
+  parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: 2018,
-    sourceType: 'module'
+    sourceType: 'module',
+    project: './tsconfig.json',
+    // allows eslint from any dir
+    tsconfigRootDir: __dirname,
+    sourceType: 'module',
```
dfreeman, line 10: Looks like this key wound up in there twice

addon/-private/class/modifier.ts
```diff
@@ -7,7 +7,9 @@ import { ModifierArgs } from "ember-modifier/-private/interfaces";
 // SAFETY: we're lying to the compiler here about the result of `symbol(...)`,
 // and this only holds *because* the implementation of `symbol` guarantees
 // sufficient uniqueness for this to make sense.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const DESTROYING: unique symbol = symbol("destroying") as any;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const DESTROYED: unique symbol = symbol("destroyed") as any;
```
dfreeman, line 13: Could these be e.g. `export const DESTROYED = symbol(“destroyed”) as unique symbol` to avoid the linter directive comments? (Or maybe even update the return type of `symbol` with a comment on the correctness there?)

addon/-private/class/modifier.ts
```diff
@@ -58,25 +62,33 @@ export default class ClassBasedModifier<
    * Called when the modifier is installed **and** anytime the arguments are
    * updated.
    */
-  didReceiveArguments() {}
+  // Disabled because this exists to be a no-op (as well as for docs).
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  didReceiveArguments(): void {}
```
dfreeman, line 67: I usually put a comment in the function body explaining why it’s empty to avoid having to explicitly disable this rule

.eslintrc.js
rwjblue: Whoops, sorry probably did my suggestion wrong 😢

.eslintrc.js
```diff
@@ -45,9 +55,26 @@ module.exports = {
         node: true
       },
       plugins: ['node'],
+      // This is a workaround for the fact that glob overrides are not allowed
+      // to use `extends` (https://github.com/eslint/eslint/issues/8813).
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
```
rwjblue, line 60: FWIW, this is allowed now (I think I mentioned in my other review)

You can tweak this to:

```js
{
  // ...snip...
  extends: ['plugin:node/recommended'],
  rules: {
    // the actual overrides you need
  }
}
```

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "8"
+  - "10"
```
rwjblue, line 6: This probably can be reverted once rebased

.vscode/settings.json
rwjblue: I still don't like this 😡, but I guess it isn't worth pwning everyone that works here based on my "editor puristism".

package.json
```diff
@@ -66,7 +69,7 @@
     "typescript": "^3.7.5"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
+    "node": "10.* || >= 12.*"
```
rwjblue, line 72: should drop on a rebase

.vscode/settings.json
chriskrycho: FWIW, I'm entirely with you. It ticks me off every repo I have to add this to.

.eslintrc.js
chriskrycho: Fixed!

addon/-private/class/modifier.ts
chriskrycho: Ahhh, that's a great idea. Will do that!

.travis.yml
chriskrycho: Rebased it into nonexistence!

addon/-private/class/modifier.ts
chriskrycho: Unfortunately, the first one doesn't work: `as unique symbol` is not allowed syntax, actually! That was the first thing I tried, but you get:

> 'unique symbol' types are not allowed here. ts(1335)

I think doing it on `symbol` is the right move, though: just make it return `any` and comment about safety there. (Especially since we'll be dropping it shortly.)

.eslintrc.js
chriskrycho: Integrated this, plus our discussion in Discord DMs about the TS config – I like that and I think we'll probably upstream it into the ember-cli-typescript docs (or even blueprint?) for this that I want to supply.

.travis.yml
```diff
@@ -56,6 +58,11 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-canary
     - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
 
+    - stage: "Type tests"
+      env: EMBER_TRY_SCENARIO=typescript-3.6
+    - env: EMBER_TRY_SCENARIO=typescript-3.7
+    - env: EMBER_TRY_SCENARIO=typescript-next
```
chriskrycho, line 64: This doesn't work because we have a separate config file, so it's not invoked correctly. Options:

- pull these back into the main ember-try config file and just specify the `command` argument there
- split them out in the Travis config somehow?

My inclination is the former; it's slightly more verbose but should get the job done just fine and with no tweaking to Travis config other than including the scenarios.

chriskrycho: The one upside to the latter is that it's possible to run (e.g. locally) `ember try:each --config-path='./config/ember-try-typescript.js` and see that output easily.

chriskrycho:
We have a working and passing CI setup!

config/ember-try.js
```diff
@@ -1,62 +1,68 @@
...
 
-const getChannelURL = require('ember-source-channel-url');
+const getChannelURL = require("ember-source-channel-url");
+const typeTests = require("./ember-try-typescript");
 
-module.exports = async function() {
+const typeScriptScenarios = typeTests.scenarios.map((s) => ({
+  ...s,
+  command: typeTests.command,
+}));
```
chriskrycho, line 9: I'm not *in love* with this but it works while allowing the test config itself to remain simple *and* letting users do `ember try:each --config-path=./config/ember-try-typescript.json` to run *just* the type tests locally.

.travis.yml
chriskrycho: I worked around this by handling it with some JS (see [below](https://github.com/ember-modifier/ember-modifier/pull/30/files#diff-d20f4baf61c4397c467d6218d0ba6a04R6-R9)).

.eslintrc.js
```diff
@@ -45,9 +48,39 @@ module.exports = {
...
       plugins: ['node'],
-      rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
+      extends: ['plugin:node/recommended'],
+      rules: {
         // add your custom rules and overrides for node files here
-      })
-    }
+        "@typescript-eslint/explicit-function-return-type": ["off"],
+        "@typescript-eslint/no-var-requires": ["off"],
+        "@typescript-eslint/camelcase": ["off"]
```
rwjblue, line 56: I think these can be removed now, `@typescript-eslint` is only used on `*.ts` files and this set of overrides doesn't match any `.ts` files.

config/ember-try.js
rwjblue: this seems fine (especially while we iterate)

config/ember-try.js
```diff
@@ -1,62 +1,68 @@
...
+
+module.exports = async function () {
   return {
     useYarn: true,
     scenarios: [
       {
-        name: 'ember-lts-2.18',
+        name: "ember-lts-2.18",
```
rwjblue, line 16: Lets avoid rewriting this file with prettier + double quotes (only to change subsequently to single quotes in a follow up).

.travis.yml
rwjblue: We could also make a single job that does (note: I don't _particularly_ care, just pointing it out as an easy alternative):

```yml
- stage: "Type Tests"
  script: yarn test:types
```

.eslintrc.js
chriskrycho: Ah, thank you!

config/ember-try.js
chriskrycho: Dagnabbit, I though I’d avoided committing any of those. Reverting.

.travis.yml
chriskrycho: Is there any difference in performance? I don’t have a preference myself either, other than wanting to come up with something that’s both easy to maintain and easy to extract for other addons to use as well.

config/ember-try.js
chriskrycho: Fixed, force pushed.

.travis.yml
chriskrycho: I just realized what you were actually saying; there would definitely not be. I prefer this because it means there's one fewer thing to change when changing supported TS versions.

chriskrycho: On reflection, I changed my mind here because we want to allow one of the environments to fail: `typescript@next`. Related, though: is there a way to have Travis+GitHub notify code owners when an allowed-to-fail build fails? It'd be great if we got notice as early as possible.

PR #38, opened 2020-05-22
Simplify the teardown APIs

### LINKED ISSUE #29: Proposal: change `element` availability
Right now, the class-based modifier specifies that the element is not available during the constructor or during `willDestroy`. The first of these is non-negotiable within the constraints of [the current modifier manager spec](https://github.com/emberjs/rfcs/blob/82305c4e502183578dcd371bc14ed7092c4f9852/text/0373-Element-Modifier-Managers.md#capabilities), and seems unlikely to change. However, the second is a *design choice*—I suspect it was made out of a worry that it would cause a memory leak, or perhaps to mirror the API from classic Ember components. Regardless, the modifier manager *forces* this:

https://github.com/ember-modifier/ember-modifier/blob/14d34b0edff357e089463c32e8b41a2b8f5823aa/addon/-private/class/modifier-manager.js#L29-L45

After discussing this with @rwjblue and @pzuraq, I propose we stop removing the element during `willDestroy`: it will be garbage collected properly during teardown of the modifier and element, and having this restriction in place substantially increases the complexity of both the docs and the types (see discussion on #23 for details on the latter), while providing no concrete benefit.

This does raise a question about the utility of having *both* `willRemove` and `willDestroy`. If we leave `this.element` in place for `willDestroy`, should we just deprecate `willRemove` as extraneous, and plan to remove it in a later major?

chriskrycho:
Resolves #29

- make `this.element` available during `willDestroy`
- deprecate the `willRemove` hook, since it's redundant
- remove the now-unnecessary `InTeardown` utility type
- update the docs

Note that this is *not* a breaking change: the previous behavior will continue to work as expected. The deprecation introduced here will remain *until* a 3.0, whatever the trigger for that release may be (e.g. when Node 10 LTS reaches end of life).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,6 @@ is based on their work, and wouldn't have been possible without them.</i>
     - [API](#api)
       - [Lifecycle Summary](#lifecycle-summary)
 - [TypeScript](#typescript)
-  - [Type safety tradeoffs](#type-safety-tradeoffs)
-  - [Lifecycle hooks and types](#lifecycle-hooks-and-types)
   - [Examples with TypeScript](#examples-with-typescript)
     - [Functional modifier](#functional-modifier)
     - [Class-based](#class-based)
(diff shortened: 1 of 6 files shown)


rwjblue:
@chriskrycho - FYI - Going to leave this open for a bit so @pzuraq has a chance to review.

chriskrycho:
Seems good to me!

andreyfel:
Hey guys! What is the status of this? Is this PR waiting for @pzuraq's review?

PR #41, opened 2020-06-04
TS: upgrade to 3.9; implement RFC for type stability

chriskrycho:
In addition to upgrading the supported TS version, implements the *technical* recommendations of typed-ember/ember-cli-typescript#1158. A later PR will add supported-version policy to the README.

Specific changes made here:

- Upgrade TypeScript to 3.9.
- Add deps to support downleveling, and add downlevel-dts config to package.json. (This has some limitations, but is preferable to not correctly supporting older versions at all.)
- Configure prepublish hook to generate downleveled type definitions as  well as the definitions for the current version in use by the addon.


package.json
```diff
@@ -20,13 +20,18 @@
...
     "start": "ember serve",
     "test": "ember test",
+    "test:ts": "tsc --noEmit",
     "test:types": "ember try:each --config-path=ember-try-typescript.js",
     "test:all": "ember try:each",
-    "prepublishOnly": "ember ts:precompile",
-    "postpublish": "ember ts:clean"
+    "prepublish:types": "ember ts:precompile",
+    "prepublish:downlevel": "downlevel-dts . ts3.4",
+    "prepublishOnly": "run-s prepublish:types prepublish:downlevel",
```
rwjblue, line 30: FWIW, I generally prefer the longer `npm-run-all` name (which is serial by default) because:

1) the name is an exact match for a package listed in this very file
2) searching for that name turns up immediate results pointing to the package

package.json
```diff
@@ -20,13 +20,18 @@
     "lint:js:fix": "eslint . --fix",
     "lint": "npm-run-all --aggregate-output --continue-on-error --parallel 'lint:!(fix)'",
     "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
-    "test:ts": "tsc --noEmit",
     "start": "ember serve",
     "test": "ember test",
+    "test:ts": "tsc --noEmit",
```
rwjblue, line 25: I know this is a sorting fix, but I'm curious to know your thoughts on naming this `lint:ts` instead of `test:ts`?

Specifically, this sort of type checking _feels_ like a lint thing and fits with other lint related tooling. I'd also prefer to not even run tests if `yarn lint` itself fails.

package.json
```diff
@@ -105,5 +112,11 @@
       "release": true,
       "tokenRef": "GITHUB_AUTH"
     }
+  },
+  "types": "index.d.ts",
+  "typeVersions": {
```
rwjblue, line 117: :sob: wish I could check in types for .json files :sob:

Can you link me to the docs for this?

package.json
chriskrycho: There's an interesting semantic argument here that is probably worth hashing out as a community—maybe on [the RFC](https://github.com/typed-ember/ember-cli-typescript/pull/1158). Whether we encourage people to think of TS checks as *lints* or as *tests* gets at a pretty deep difference in how people approach it. Is a compiler error comparable to an ESLint error—i.e. something that could just be a matter of opinion? Or does it represent a *bug* the way a failing test does?

package.json
chriskrycho: Ah, so `npm-run-all --serial`? 👍

chriskrycho: Oh, reading fail—I didn't realize it was serial by default. 👍

package.json
chriskrycho: Yeah, at a minimum schema validation would help. Docs are [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions).

…and I'm glad you asked about this, because the key is *actually* `typesVersions`; I just typo'd it. 🤦‍♂️

chriskrycho:
Oh, that's a great catch on `prepublish`. We *should* use it in `prepublish` to guarantee it's there, but we absolutely should run our CI tests against it as well. I'll update this *and* push an update to the RFC there.

package.json
rwjblue: Some theory arguments first (though these aren't my main focus/point):

* Linting errors represent bugs just as clearly as failing tests (obviously there are non-bug related lints e.g. prettier, but overall the things in most sets of `recommended` / `plugin:node/recommended` / etc configs **do** represent bugs). The primary differences are: linting is fast whereas tests are slow.
* This is essentially fallout of the ember-cli-typescript teams decision to use Babel as the primary mechanism for transpiling `.ts` files (note: I'm not saying this was wrong or bad in anyway). Babel literally **does not care** about the correctness of the TypeScript files, it merely removes the type annotations and happily compiles for runtime completely invalid TypeScript code. This effectively makes this category of errors no different than the category of errors that eslint's recommended checks against.

Now for the practical (and to me the primary) point:

Doing this as part of `yarn lint` means that TS failures **prevent** a test run. This might seem contrary to the line of thinking you had above (where you were logically arguing that TS failures are more important so they are basically test failures), but in practical terms `yarn test` (here and in all addons by default) runs linting **first**. The goal here is to identify _fundamental_ problems before wasting developer or CI time to run tests.

package.json
rwjblue: Awesome, thank you!

package.json
chriskrycho: Yep, and to be clear I'm not necessarily making an argument about which we should pick here, and if the best way to get to fast-failure is just to use the existing lint infrastructure, that may be absolutely the right thing to do—but I *do* have concerns about the *language* of "linting" and what it implies. I'm speaking there from experience over the last couple years helping folks with TS things: while I very much agree with you about what lint errors represent, a *lot* of people don't, and when those same people treat *type errors* as *lints*, it encourages them to just hand wave them if they don't like/disagree with/think they know better than the error.

I also think you're correct philosophically about the relationship between Babel + TS here! Moreover, the concerns I have about this in the big picture are not a concern for *this* repo, because there's shared philosophy here: lint errors *are errors*, whatever their source.

I do wonder whether we might want to use a name like `check` instead of `lint` more broadly. 🤔 

I'm going to add this as an unresolved question on the RFC, and would mildly prefer we discuss there with *this* as a motivating discussion source for it.

In the meantime, I'm happy enough to switch it to `lint` here for now in the interest of landing this; we can always tweak it later!

chriskrycho: Related: we have `test:types` currently, and that's run during the CI run as part of the *test* suite. I think we want that to *remain* a "test" thing that's part of CI only, *not* treat it as a lint step, because it's *not* a fast-fail operation. Thoughts?

rwjblue: > I'm speaking there from experience over the last couple years helping folks with TS things: while I very much agree with you about what lint errors represent, a lot of people don't, and when those same people treat type errors as lints, it encourages them to just hand wave them if they don't like/disagree with/think they know better than the error.

I have definitely seen this as well, and I do agree that people are *way* to flippant about linting errors (but IMHO this is because they are using stupid/useless linting rules that just cause pain without justification 😡 ). 

> I'm going to add this as an unresolved question on the RFC, and would mildly prefer we discuss there with this as a motivating discussion source for it.

:+1: - Will move convo over to the RFC thread.

> In the meantime, I'm happy enough to switch it to lint here for now in the interest of landing this; we can always tweak it later!

👍

chriskrycho: - [Added this as an unresolved question on the RFC](https://github.com/typed-ember/ember-cli-typescript/pull/1158/commits/ec8f777c9b055168fb52466d0a8c60204b513a38), and am about to push the change here (along with the `typesVersions` fix).
- Totally agreed re: stupid/useless linting rules and their impact on people's habits. 😭

rwjblue: > Related: we have test:types currently, and that's run during the CI run as part of the test suite. I think we want that to remain a "test" thing that's part of CI only, not treat it as a lint step, because it's not a fast-fail operation. Thoughts?

Yes, totally agree. It is running _much_ more than a "normal" `tsc --noEmit`, as it is testing each supported `tsc` version.

chriskrycho:
Agreed, working on changing that now!

package.json
```diff
@@ -23,11 +23,13 @@
     "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
     "start": "ember serve",
     "test": "ember test",
-    "test:types": "ember try:each --config-path=ember-try-typescript.js",
+    "pretest": "npm-run-all prepublish:*",
+    "posttest": "npm run clean",
+    "test:types": "ember try:each --config-path=./config/ember-try-typescript.js",
```
chriskrycho, line 28: I *believe* this config should work correctly in terms of generating the types we need. It will run `downlevel-dts` *prior* to running the `ember-try` suite for tests—both when running `test:types` and when running the full `ember try:each` suite in `test:all`—so users can verify directly *and* it will be correctly configured on CI runs, so that we actually check the type tests against the publicly generated types.

However, it doesn't currently actually _invoke_ the type tests at `types/modifier-test.ts`—they're not emitted, so they won't be checked!—so we need to change *that* configuration in some way. 🤔 

@dfreeman @jamescdavis do you have thoughts here?

chriskrycho: The only things I've come up with are to *manually* copy the tests into the target directory from down-leveling, or to put the type tests in a spot where they get included automatically by `ember ts:precompile` (probably suggesting a need to change the behavior of `ember-cli-typescript` to support it).

chriskrycho: I think I came up with something reasonable; see 21c57a2

PR #46, opened 2020-07-28
Use ember-cli-htmlbars for inline precompilation.

rwjblue:
As of `ember-cli-htmlbars@4` inline precompilation is provided natively by `ember-cli-htmlbars` (and you no longer need `ember-cli-htmlbars-inline-precompile` as a dep/devDep). This updates both our tests and the tests generated by our blueprints to use `ember-cli-htmlbars` directly.

### DIFF
--- a/blueprints/modifier-test/index.js
+++ b/blueprints/modifier-test/index.js
@@ -1,8 +1,6 @@
 'use strict';
 
 const stringUtils = require('ember-cli-string-utils');
-const isPackageMissing = require('ember-cli-is-package-missing');
-
 const useTestFrameworkDetector = require('../test-framework-detector');
 
 module.exports = useTestFrameworkDetector({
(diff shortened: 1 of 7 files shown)


blueprints/modifier-test/index.js
```diff
@@ -38,16 +38,4 @@ module.exports = useTestFrameworkDetector({
...
-    if (
-      !options.dryRun &&
-      options.testType === 'integration' &&
-      isPackageMissing(this, 'ember-cli-htmlbars-inline-precompile')
-    ) {
-      return this.addPackagesToProject([
-        { name: 'ember-cli-htmlbars-inline-precompile', target: '^0.3.1' },
-      ]);
-    }
-  },
```
rwjblue, line 52: No longer needed, since we can reasonably assume that the host app/addon has `ember-cli-htmlbars` itself installed already.

PR #50, opened 2020-07-30
Avoid issuing a deprecation during destruction on Ember 3.20+

rwjblue:
Ideally, this could have just backported the changes from 02ebf5cbb5369506ec23c67da692a659cba6e37c, but unfortunately v1.x of ember-modifier still supports Node 8 (and `ember-destroyable-polyfill` does not).

Instead, this avoids the deprecation (by avoiding the methods that issue the deprecation on the affected versions of Ember). It also ensures that on Ember 3.20 (and eventually users of the non-polyfilled final implementation in Ember 3.22+) _other_ users of ember-destroyable-polyfill **can** use `registerDestructor` and reliably have that invoked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
 branches:
   only:
     - master
+    - v1.x
     # npm version tags
     - /^v\d+\.\d+\.\d+/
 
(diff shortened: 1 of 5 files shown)


rwjblue:
This confirms that on current `release` channel (3.20) there are no deprecations:

https://travis-ci.com/github/ember-modifier/ember-modifier/jobs/366877088

PR #56, opened 2020-08-14
config: update TS and TS test matrix

chriskrycho:
- Use *latest* TypeScript 3.9 for developing the addon (has a few compiler bug fixes in it).
- Update the ember-try config to include 3.9 explicitly in the test matrix, while maintaining support for 3.6, 3.7 and `next`.
- Use the ember-cli-typescript 4.0 release candidate: combined with latest ember-cli-babel, it makes everything work nicely on TS 3.9 and 4.0. We're not currently using the features where it matters, but it keeps things tidy.
- Create a shared `tsconfig.json` to be used by both the *current* and *next* version of TypeScript, which does *not* use down-leveled types.

Previously, we were depending on everything to keep type-checking just by dint of the normal CI pass, but weren't explicitly checking the type tests against 3.9. Whoops. 😬 This was "fine" (as in this-is-fine.gif) because we were checking against TS@next, but we need to make sure we run the type tests against the *current* TS version as well as earlier versions and `next`.


yarn.lock
```diff
@@ -6022,6 +6051,8 @@ execa@^3.0.0:
     signal-exit "^3.0.2"
     strip-final-newline "^2.0.0"
 
+=======
+>>>>>>> f3699f8... config: update TS and TS test matrix
```
Turbo87, line 6055: @chriskrycho looks like merge conflicts to me 😉

chriskrycho: Ahhhh, good catch, thank you.

package.json
```diff
@@ -40,7 +40,7 @@
     "ember-cli-babel": "^7.22.1",
     "ember-cli-normalize-entity-name": "^1.0.0",
     "ember-cli-string-utils": "^1.1.0",
-    "ember-cli-typescript": "^3.1.3",
+    "ember-cli-typescript": "^4.0.0-rc.1",
```
rwjblue, line 43: I believe 4.0.0 final is out, can you update?

package.json
```diff
@@ -125,7 +125,9 @@
   "types": "index.d.ts",
   "typesVersions": {
     "<3.9": {
-      "*": "ts3.4/*"
+      "*": [
+        "ts3.4/*"
+      ]
```
rwjblue, line 130: Was the old config wrong?

package.json
chriskrycho: Will do!

package.json
chriskrycho: It was not; what it was is *easy to mess up*. Doing it this way Kees things correct more easily. I plan to document this in the SemVer for Addon Types RFC and hopefully get it codified into tooling early October.

package.json
chriskrycho: Updated, and for TS 4.0 as well.

chriskrycho:
Updated this to latest; it's now failing with an ESLint parse error:

```
/Users/ckrycho/dev/ember-modifier/ember-modifier/tests/integration/modifiers/functional-modifier-test.ts
  0:0  error  Parsing error: Cannot read property 'map' of undefined

/Users/ckrycho/dev/ember-modifier/ember-modifier/type-tests/modifier-test.ts
  0:0  error  Parsing error: Cannot read property 'map' of undefined
```

Anybody seen this recently? Otherwise I'll go poking later.

rwjblue:
Yikes, no I've not seen that before.

chriskrycho:
_le sigh_. I’ll see if I can get to the bottom of it later this week or early next week.

chriskrycho:
Will circle back to this late next week and see if I can get it sorted. Possibly just needs a proper rebase.

chriskrycho:
Superseded by #97.

PR #57, opened 2020-08-14
Update documentation to use `assert` from `@ember/debug` for type narrowing

chriskrycho:
When I wrote the previous iteration of the docs, we did not yet have a version of Ember's `assert` which used the TS 3.7 `asserts` operator. We do now, so we can safely recommend people use that for type narrowing. :tada:

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,26 +698,25 @@ True type safety requires runtime checking, since templates are not currently ty
 
 ```ts
 import Modifier from 'ember-modifier';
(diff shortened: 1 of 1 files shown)


buschtoens:
omg I did not know the types have it now 😍

chriskrycho:
As of a huge push I made a month or two ago! It’s SO MUCH BETTER. M

Turbo87 (approved):

PR #63, opened 2020-10-15
feat(modifier-managers): use 3.22 capabilities

NullVoxPopuli:
BREAKING CHANGE: destruction behavior changed to use new args proxy from
3.22

Resolves the issue 
    described by: https://github.com/emberjs/ember.js/issues/19162
    fixed by: https://github.com/emberjs/ember.js/pull/19163


.travis.yml
```diff
@@ -47,11 +47,6 @@ jobs:
 
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.4
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.8
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.12
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.16
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.20
```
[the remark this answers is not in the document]
NullVoxPopuli: :p 
thanks for the compatibility helpers tip on the other PR :D <3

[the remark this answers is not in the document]
rwjblue: NP! This repo will have to handle it a bit differently than ember-render-modifiers though.

tests/integration/modifiers/class-modifier-test.ts
```diff
@@ -472,11 +472,11 @@ module('Integration | Modifier Manager | class-based modifier', function (
         }
 
         didReceiveArguments(): void {
-          callback('didReceiveArguments', this);
+          callback('didReceiveArguments', this, this.args.named.foo)
```
NullVoxPopuli, line 475: this is incorrect type-wise, but without them there are two additional errors 🙃

addon/-private/functional/modifier-manager.ts
```diff
@@ -28,15 +29,25 @@ function setup(
...
 
-  createModifier(factory: Factory): FunctionalModifier {
+  createModifier(...[factoryOrClass]: CreateModifierArgs): FunctionalModifier {
     // This looks superfluous, but this is creating a new instance
     // of a function -- this is important so that each instance of the
     // created modifier can have its own state which is stored in
     // the MODIFIER_ELEMENTS and MODIFIER_TEARDOWNS WeakMaps
-    return (...args) => factory.class(...args);
+    return (...args) => {
+      if ('class' in factoryOrClass) {
```
NullVoxPopuli, line 45: should this also be wrapped in a `if ember@3.13`? do the comparability helpers get burned in at build time? (like DEBUG?)

NullVoxPopuli:
green! yay!

type-tests/modifier-test.ts
```diff
@@ -3,13 +3,19 @@ import { expectTypeOf } from 'expect-type';
...
       positional: unknown[],
       named: Record<string, unknown>
-    ) => (() => unknown) | void
+    ) =>
+      | ((
+          element: Element,
+          positional: unknown[],
+          named: Record<string, unknown>
+        ) => unknown)
+      | void
```
chriskrycho, line 18: A couple issues here:

1. This should *not* be `toMatchTypeOf`; `toEqualTypeOf` is stricter in ways we *very much need*. Specifically: [making it so we can't incorrectly broaden the type, e.g. to `any`][playground].
2. As far as I can tell, the changes to functional modifiers should not have changed the public interface here: why is it now expected for the callback to possibly return *another* callback with the same signature?

[playground]: https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwDkaG2AtDAcnQNwBQ3bGcALIQAJsGLBkUAIJQA5gGc4AXkTc4GuADsAhiGQiAXHABKWaCIA8CmFGBa5AGjgBXLQGstEAO5aAfI7qmpAKwDDAELoANsZunj5aANoAuoFkPNyMWLjsJJbCYhJSsop+ABQAlAB0MBCCOjCYABb4RMSWOlp45RWZ6NmtyHkF4pIy8go9NRAAogCOLjpRg3md3ZXcQA

chriskrycho (changes requested):

type-tests/modifier-test.ts
NullVoxPopuli: I don't know what's different between my work machine and my home machine (what I was using when I made this change), but on my work machine, _all_ type tests fail for every scenario (12 errors on each of TS 3.6, 3.7, and next).

Due to the uncertainty (and vague errors), I've reverted this change, and we'll let travis decide what's going on.

Note: two kinds of vague errors:
 - `Expected 2 arguments, but got 0`
 - `An argument for 'expected' was not provided`

chriskrycho: I've just pulled it down and am testing, will see if I see the same.

chriskrycho: Ah, so: the problem is actually with our *setup* for the down-leveled tests. It's not clear to me why CI isn't catching this, but: we're actually just not resolving the import for the module at all! The first error in the logs:

```
type-tests/modifier-test.ts:3:50 - error TS2307: Cannot find module 'ember-modifier'.

3 import Modifier, { modifier, ModifierArgs } from 'ember-modifier';
                                                   ~~~~~~~~~~~~~~~~
```

All the other errors are just the result of the fact that the types imported from `ember-modifier` *don't resolve to anything*! I'll poke at this further.

chriskrycho: Aaaaand, I figured it out. In all my testing when setting this up, I assumed that we'd be running `yarn prepublishOnly` for invoking these, and I believe that is indeed what's happening on CI, which is why it's passing there. It's easy to forget to do that locally, so I'll think on (and cc. @dfreeman and @jamescdavis) the right design for making it easy to get the ember-try scenarios correct. (It may be as simple as making the command be `prepublishOnly && ...`) But when you do that, the test scenarios behave as expected. So as you're aiming to iterate here, now you know what to do!

NullVoxPopuli: thanks for the tip!
now there is only one error for each scenario:
```
=== Scenario: typescript-3.6 ===================================================

type-tests/modifier-test.ts:6:24 - error TS2554: Expected 2 arguments, but got 0.

  6 expectTypeOf(modifier).toEqualTypeOf<
                           ~~~~~~~~~~~~~~
  7   (
    ~~~
... 
 13   ) => unknown
    ~~~~~~~~~~~~~~
 14 >();
    ~~~

  node_modules/expect-type/dist/index.d.ts:66:31
    66     toEqualTypeOf: <Expected>(expected?: Expected, ...MISMATCH: MismatchArgs<Equal<Actual, Expected>, B>) => true;
                                     ~~~~~~~~~~~~~~~~~~~
    An argument for 'expected' was not provided.


Found 1 error.


Result: false
---

```

chriskrycho: That’s because that type is wrong. Like I said above, I do not see anything in the implementation indicating why you added it. What were you going for?

NullVoxPopuli: I don't know who to believe anymore:

image

NullVoxPopuli: oh its no longer in the GH diff.

but yeah. the error I get are _with_ the original code. I changed nothing now. anywho

chriskrycho: Seems to be passing both on CI and locally for me now that you’re reverted. 🤷‍♂️ 

Please confirm you have tried this:

```sh
$ yarn prepublishOnly
$ ember try:one typescript-3.6
```

NullVoxPopuli:
```
Scenario typescript-3.6: SUCCESS
```
guess I forgot prepublishOnly after I reverted the change

andreyfel:
Hey guys! What is the status of this? Will it be merged?

alexlafroscia:
I tested out this branch locally as a potential fix for #74 and it seems like it resolves my problems with argument values not always being what you might expect.

I hate to do this because I know maintainer time is really precious, but could you take a look at whether this is merge-ready @rwjblue ? It would be really nice to get this fix out to the community!

jelhan:
Ember introduced a deprecation if using modifiers with capabilities less than 3.22 in canary:

> Versions of modifier manager capabilities prior to 3.22 have been deprecated. You must update to the 3.22 capabilities.

Would be great to get this one before the deprecation is included in a stable release.

alexlafroscia:
Now that the warning that @jelhan mentioned is hitting a publicly-available version of Ember, it would be really, _really_ nice to land this PR 😬

jelhan:
> BREAKING CHANGE: destruction behavior changed to use new args proxy from 3.22

@NullVoxPopuli In which sense is this a breaking change? What changes to a consuming addon are needed? Would be great to get some guidance. 🙃

pzuraq:
The autotracking semantics of the new args proxy are what changed. Arguments are no longer eagerly consumed - you must use an argument for it to be part of autotracking.

The conclusion we had at the last meeting where we discussed this was that we do not need to make this a breaking change. Instead, the modifier manager can continue to consume all arg values eagerly itself, and we can figure out the transition path for the breaking change later on, so we can get this merged and then figure out how to have the whole ecosystem convert sometime in the future.

alexlafroscia:
Thanks for the context @pzuraq!

addon/-private/class/modifier-manager.ts
```diff
@@ -10,16 +12,40 @@ function destroyModifier(modifier: ClassBasedModifier): void {
...
-    const Modifier = factory.class;
+    const [factoryOrClass, args] = createModifierArgs;
+
+    let Modifier: typeof ClassBasedModifier;
+
+    if (gte('3.22.0')) {
+      assert(
+        'expected factory to be a class reference',
+        !('class' in factoryOrClass)
+      );
```
rwjblue, line 38: I don't fully grok this assertion. What is the point of it? It seems perfectly reasonable to have a Modifier that looks like:

```js
export default class extends Modifier {
  static get class() {
    // stuff!
  }
}
```

Its kinda dumb looking, but it isn't invalid.

---

These two assertions seem like "testing the framework" to me, and I think they could be removed..

addon/-private/functional/modifier-manager.ts
rwjblue: Yes, this should also be guarded (to only be done when we know it needs to be).

sandydoo:
> The conclusion we had at the last meeting where we discussed this was that we do not need to make this a breaking change. Instead, the modifier manager can continue to consume all arg values eagerly itself, and we can figure out the transition path for the breaking change later on, so we can get this merged and then figure out how to have the whole ecosystem convert sometime in the future.

@pzuraq, so this PR should be rewritten to consume all the args, right? `ember-render-modifiers` is also blocked on the same issue with `did-update`.

pzuraq:
Correct, I haven't had time to push this through but would really appreciate the help!

PR #77, opened 2021-04-10
migrate from TravisCI to GitHub Actions

jelhan:
GitHub Actions configuration was created with [create-github-actions-setup-for-ember-addon](https://github.com/jelhan/create-github-actions-setup-for-ember-addon). Haven't noticed anything special in existing TravisCI configuration. So it should work out of the box. But please double check before merging.


pzuraq:
It seems like the action wasn't triggered for this PR, which is a bit concerning. Do we need to enable actions somehow? Normally they Just Work™️ in my experience

rwjblue (approved):

rwjblue:
Lets see what happens...

jelhan:
> Lets see what happens...

Thanks for the trust. :sweat_smile:

Merging triggered this CI pipeline for `master` branch: https://github.com/ember-modifier/ember-modifier/runs/2326155181 Looks good so far, isn't it?

rwjblue:
Yep, seems like things are good!

PR #22, opened 2020-05-04
docs(README): fix arguments destructuring

MrChocolatine:
Fix #21

Issue #21, opened 2020-05-02
README - Class modifier - Arguments destruct incorrect


MrChocolatine:
I am not quite sure the arguments can be destructured from the Class itself :
https://github.com/ember-modifier/ember-modifier/blob/96f64dfb5036d99a3b4648a71db071077e8f60de/README.md#L181

I guess it should be the same as :
https://github.com/ember-modifier/ember-modifier/blob/96f64dfb5036d99a3b4648a71db071077e8f60de/README.md#L171

pzuraq:
Yep, that looks correct to me. PR welcome here, I'll see if I can get it fixed soon but pretty busy at the moment

MrChocolatine:
I completely forgot.

joshkg:
> `this` refers to the Class, not an Array, therefore Array destructuring is not what we want.

@MrChocolatine Agree that it should have been object destructuring, but the question is really about the logic -- I believe the reason we were targeting `this` was to check for previously set `event` and `handler` properties as noted above

MrChocolatine:
> > `this` refers to the Class, not an Array, therefore Array destructuring is not what we want.
> 
> @MrChocolatine Agree that it should have been object destructuring, but the question is really about the logic -- I believe the reason we were targeting `this` was to check for previously set `event` and `handler` properties as noted above

I see, sorry I misunderstood your message then!

That is actually a fair point!

1. The modifier is called a first time, everything runs correctly.
1. The arguments are updated, `didReceiveArguments()` gets called.
1. `removeEventListener()` gets called, we might actually prefer to use/remove the `event` and `handler` already stored on `this`
1. `addEventListener()` will run using the updated arguments.

joshkg:
@MrChocolatine Right! So I believe it will help clarify this example to either remove the logic that is storing the `event` and `handler` on the instance, and always destructure them from `this.args.positional`, OR retain the logic that is storing them and update the `removeEventListener` to destructure `event` and `handler` from `this`.

MrChocolatine:
> retain the logic that is storing them and update the `removeEventListener` to destructure `event` and `handler` from `this`.

Personally, I would vote for this.

MrChocolatine:
Let's give it a try:
https://github.com/ember-modifier/ember-modifier/pull/90

Ping @joshkg and @pzuraq .

PR #90, opened 2021-07-02
Improve logic in `removeEventListener()` from examples

MrChocolatine:
Improve logic in `removeEventListener()` from examples by reusing the properties `event` and `handler` stored in the class instead of using `this.args.positional` (which could potentially give us new and different parameters from what has been stored/set up before).

Relates to this comment:
https://github.com/ember-modifier/ember-modifier/issues/21#issuecomment-853204785

> * The modifier is called a first time, everything runs correctly.
> * The arguments are updated, `didReceiveArguments()` gets called.
> * `removeEventListener()` gets called, we might actually prefer to use/remove the `event` and `handler` already stored on `this`
> * `addEventListener()` will run using the updated arguments.

Undo and improve what has been done in:
https://github.com/ember-modifier/ember-modifier/pull/22

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,7 +352,7 @@ export default class OnModifier extends Modifier {
   }
 
   removeEventListener() {
-    let [event, handler] = this.args.positional;
+    let { event, handler } = this;
 
     if (event && handler) {
       this.element.removeEventListener(event, handler);


joshkg:
LGTM!

MrChocolatine:
Ping @pzuraq, @chriskrycho, @rwjblue :v: .

pzuraq:
Thank you!

PR #96, opened 2021-10-14
Drop Node 10 and add Node 14+

chriskrycho:
- update `engines` config
- add `volta` config
- incidentally fix some issues with test config and TS


chriskrycho:
The failures here will be unblocked by #97.

chriskrycho:
Ugh, nope, they’re going to have to land together.  😩

PR #212, opened 2022-03-21
Deprecate `willDestroy()` and `isDestroy(ing|ed)`

chriskrycho:
With the availability of the `@ember/destroyables` API, we no longer need these to be handled via lifecycle hooks. Users previously set up their own `willDestroy` hook, and it got called for them:

```ts
import Modifier from 'ember-modifier';

class MyModifier extends Modifier {
  willDestroy() {
    // ...
  }
}
```

Now, like with other classes in modern Ember/Glimmer, this can stop being a "lifecycle hook" and simply use the `registerDestructor` API:

```ts
import Modifier from 'ember-modifier';
import { registerDestructor } from '@ember/destroyable';

class MyModifier extends Modifier {
  constructor(owner, args) {
    super(owner, args);
    registerDestructor(this, () => {
      // ...
    });
  }
}
```

Similarly, instead of checking the `.isDestroyed` or `.isDestroying` properties on the instance, users can import the correspondingly-named functions from `@ember/destroyables`, e.g. `isDestroyed(instance)`.

### DIFF
--- a/addon/-private/class/modifier.ts
+++ b/addon/-private/class/modifier.ts
@@ -3,6 +3,26 @@ import { setModifierManager } from '@ember/modifier';
 import Manager from './modifier-manager';
 import { isDestroying, isDestroyed } from '@ember/destroyable';
 import { ElementFor, ArgsFor, DefaultSignature } from '../signature';
+import { deprecate } from '@ember/debug';
+
+function deprecateForDestroyables<S>(
+  name: 'willDestroy' | 'isDestroying' | 'isDestroyed',
+  instance: ClassBasedModifier<S>
+): void {
+  deprecate(
+    `Modifier.${name} is deprecated`,
+    instance[name] !== ClassBasedModifier.prototype[name],
+    {
+      id: 'ember-modifier.use-destroyables',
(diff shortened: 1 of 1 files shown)


addon/-private/class/modifier.ts
```diff
@@ -40,6 +60,10 @@ export default class ClassBasedModifier<S = DefaultSignature> {
   constructor(owner: unknown, args: ArgsFor<S>) {
     setOwner(this, owner);
     this.args = args;
+
+    deprecateForDestroyables('willDestroy', this);
+    deprecateForDestroyables('isDestroying', this);
+    deprecateForDestroyables('isDestroyed', this);
```
rwjblue, line 66: I think it's very very unlikely that someone overrides the `isDestroying`/`isDestroyed` part.

rwjblue (approved):
LGTM

addon/-private/class/modifier.ts
chriskrycho: Yeah, I assume the same, but if someone *can* override those things... 😅

PR #217, opened 2022-03-22
Introduce new `modify` hook

chriskrycho:
This hook is the future of the class-based modifier API; it replaces *all* the non-destruction lifecycle hooks. The new API closely mirrors both the existing function-based modifier design and the design of class-based helpers in Ember:

```ts
import Modifier from 'ember-modifier';

export default class MyModifier extends Modifier {
  modify(element, positional, named) {
    // ...
  }
}
```

This API shift results in several significant improvements compared to the previous API:

1.  The signature of `Modifier.modify` matches the signature of a function-based signature. Both receive the element and both positional and named arguments directly, and in the same positions. This means that transitioning from a function- to a class-based helper is as simple as using the callback passed to `modifier` as the `modify` method on the class. Users no longer need

2.  It now matches the API of a class-based `Helper`, which has only the single `compute` method. This is important for rationalizing the mental model for using them: Modifiers and helpers are *very* similar, with the key difference being that helpers are intended to be pure functions which produce values, while modifiers are intended to be impure functions which mutate (modify!) a piece of DOM and do *not* return values.

3.  This allows us to introduce the desired semantics for modifiers *without a breaking change*. Unlike the existing lifecycle hooks, `modify` does *not* eagerly consume its arguments. Instead, matching the rest of the framework, it only entangles with the tracked values it *actually uses*. This means that users should take care when migrating their existing modifiers to use the new hook.

    To maintain exact backwards compatibility, modifier authors would need to explicitly use all of the arguments passed to their modifiers. Instead, modifier authors should generally cut a new major version which has the updated, lazy semantics.

It is a hard error to use `modify` along with any of the existing (non-destruction) lifecycle hooks. This allows us to guarantee that `modify` maintains the standard auto-tracking semantics as described above.

With `modify` available, deprecate the existing lifecycle hooks; users will need to begin migrating to the new design. By introducing `modify` this way, however, users can migrate their existing modifiers piecemeal.

### DIFF
--- a/addon/-private/class/modifier-manager.ts
+++ b/addon/-private/class/modifier-manager.ts
@@ -1,14 +1,48 @@
 import { capabilities } from '@ember/modifier';
 import { gte } from 'ember-compatibility-helpers';
 import { set } from '@ember/object';
-import { destroy, registerDestructor } from '@ember/destroyable';
+import {
+  destroy,
+  registerDestructor,
+  associateDestroyableChild,
+} from '@ember/destroyable';
 
-import ClassBasedModifier from './modifier';
+import ClassBasedModifier, { _implementsModify } from './modifier';
 import { ArgsFor, ElementFor } from 'ember-modifier/-private/signature';
 import { consumeArgs, Factory, isFactory } from '../compat';
 
 function destroyModifier<S>(modifier: ClassBasedModifier<S>): void {
   modifier.willDestroy();
+}
+
+/**
+ * The state bucket used throughout the life-cycle of the modifier. Basically a
+ * state *machine*, where the framework calls us with the version we hand back
+ * to it at each phase. The two states are the two `extends` versions of this
+ * below.
+ * @internal
+ */
+interface State<S> {
+  instance: ClassBasedModifier<S>;
+  element: ElementFor<S> | null;
+}
+
+/**
+ * The `State` after calling `createModifier`, and therefore the state available
+ * at the start of `InstallModifier`.
+ * @internal
+ */
+interface CreateState<S> extends State<S> {
+  element: null;
+}
+
+/**
+ * The `State` after calling `installModifier`, and therefore the state
+ * available in all subsequent `updateModifier` calls.
+ * @internal
+ */
+interface InstalledState<S> extends State<S> {
+  element: ElementFor<S>;
 }
 
 export default class ClassBasedModifierManager<S> {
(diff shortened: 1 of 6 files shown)


addon/-private/class/modifier.ts
```diff
@@ -24,6 +30,21 @@ function deprecateForDestroyables<S>(
   );
 }
 
+/** @internal */
+export const _implementsModify = <S>(
+  instance: ClassBasedModifier<S>
+): boolean => instance.modify !== ClassBasedModifier.prototype.modify;
```
rwjblue, line 36: I know I mentioned doing it this way, but I think we should actually eagerly capture `ClassBasedModifier.prototype.modify`. If we don't then it's possible for someone to haxxxors around our check here by just modifying the prototype.

---

Note: I'm 100% ok with pushback here, maybe we just don't care if you are going to do that much dastardly haxxory...

addon/-private/class/modifier.ts
```diff
@@ -64,6 +85,55 @@ export default class ClassBasedModifier<S = DefaultSignature> {
     deprecateForDestroyables('willDestroy', this);
     deprecateForDestroyables('isDestroying', this);
     deprecateForDestroyables('isDestroyed', this);
+
+    assert(
+      'ember-modifier: You cannot implement both `run` and any of the deprecated legacy lifecycle hooks (`didInstall`, `didReceiveArguments`, and `didUpdateArguments`)',
```
rwjblue, line 90: Didn't we settle on `modify`?

addon/-private/class/modifier.ts
```diff
@@ -64,6 +85,55 @@ export default class ClassBasedModifier<S = DefaultSignature> {
...
+   *
+   *   modify(element, callback, options) {
+   *     // Gets rid of any existing connection, since we're getting called with
+   *     // a new callback or new options.
+   *     disconnect(this);
+   *
+   *     this.observer = new IntersectionObserver(callback, options);
+   *     this.observer.observe(element);
+   *
+   *     registerDestructor(this, disconnect);
```
rwjblue, line 120: FWIW, I think this would register a new destructor on every mutation (never culling/removing the prior value)

rwjblue (changes requested):

rwjblue:
@chriskrycho - This is intentionally not introducing the deprecation, right? That will come in a follow up?

addon/-private/class/modifier.ts
chriskrycho: Honestly, yeah, that’s my take. If someone does that and they get broken by it, that’s entirely their own fault. You’d have to be actively trying to haxx0rs around the deprecation (and: why? It wouldn’t actually gain you anything!) so on your own head be it at that point.

addon/-private/class/modifier.ts
chriskrycho: Haha, yep, just a cleanup failure.

addon/-private/class/modifier.ts
chriskrycho: Ah, good catch. Yeah, I’ll update the example.

chriskrycho:
Thanks for the review! Will update tomorrow, as this is all good. And yeah, I’m going to add the deprecations separately for the sake of clearer changelog and commit log.

addon/-private/class/modifier-manager.ts
```diff
@@ -21,43 +55,76 @@ export default class ClassBasedModifierManager<S> {
...
+    associateDestroyableChild(state, modifier);
     registerDestructor(modifier, destroyModifier);
 
-    return modifier;
+    return state;
   }
 
   installModifier(
-    instance: ClassBasedModifier<S>,
+    state: State<S>,
```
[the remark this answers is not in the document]
chriskrycho: Yeah, the thing this “avoids” is simply creating a new object rather than reusing the old one, but a cast would also allow that, and would be more honest (if we care about avoiding the allocation).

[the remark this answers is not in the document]
chriskrycho: Resolved via ebc9285

addon/-private/class/modifier-manager.ts
```diff
@@ -21,43 +55,76 @@ export default class ClassBasedModifierManager<S> {
...
 
-    if (gte('3.22.0')) {
+    set(instance, 'args', args); // TODO: remove aat 4.0
+
+    // The `consumeArgs()` call backwards compatibility on v3 for the deprecated
+    // legacy lifecycle hooks (`didInstall`, `didReceiveArguments`, and
+    // `didUpdateArguments`), which accidentally had eager consumption semantics
+    // prior to Ember 3.22. The new, recommended `modify` hook has the updated
+    // lazy semantics associated with normal auto-tracking.
+    const implementsModify = _implementsModify(instance);
```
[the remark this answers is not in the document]
chriskrycho: Done via ba34ba1 – great suggestion!

addon/-private/class/modifier-manager.ts
```diff
@@ -1,16 +1,50 @@
...
+ * The state bucket used throughout the life-cycle of the modifier. Basically a
+ * state *machine*, where the framework calls us with the version we hand back
+ * to it at each phase. The two states are the two `extends` versions of this
+ * below.
+ * @internal
+ */
+interface State<S> {
+  instance: ClassBasedModifier<S>;
+  element: ElementFor<S> | null;
+}
```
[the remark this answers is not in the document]
chriskrycho: Done in ba34ba1!

addon/-private/class/modifier-manager.ts
```diff
@@ -21,43 +55,76 @@ export default class ClassBasedModifierManager<S> {
...
+    state.element = element;
 
-    if (gte('3.22.0')) {
+    // The `consumeArgs()` call backwards compatibility on v3 for the deprecated
+    // legacy lifecycle hooks (`didInstall`, `didReceiveArguments`, and
+    // `didUpdateArguments`), which accidentally had eager consumption semantics
+    // prior to Ember 3.22. The new, recommended `modify` hook has the updated
+    // lazy semantics associated with normal auto-tracking.
+    const implementsModify = _implementsModify(instance);
+    if (gte('3.22.0') && !implementsModify) {
```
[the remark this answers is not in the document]
chriskrycho: Much cleaner. 👍🏼 ba34ba1

addon/-private/class/modifier-manager.ts
```diff
@@ -21,43 +55,76 @@ export default class ClassBasedModifierManager<S> {
...
-    if (gte('3.22.0')) {
+    set(instance, 'args', args); // TODO: remove aat 4.0
+
+    // The `consumeArgs()` call backwards compatibility on v3 for the deprecated
+    // legacy lifecycle hooks (`didInstall`, `didReceiveArguments`, and
+    // `didUpdateArguments`), which accidentally had eager consumption semantics
+    // prior to Ember 3.22. The new, recommended `modify` hook has the updated
+    // lazy semantics associated with normal auto-tracking.
+    const implementsModify = _implementsModify(instance);
+    if (gte('3.22.0') && !implementsModify) {
```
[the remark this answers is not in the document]
chriskrycho: ba34ba1

addon/-private/class/modifier.ts
chriskrycho: Updated in ef0a76c – let me know if that *isn’t* the right way to do it!

addon/-private/class/modifier.ts
```diff
@@ -64,6 +85,57 @@ export default class ClassBasedModifier<S = DefaultSignature> {
...
+   *     // Gets rid of any existing connection, since we're getting called with
+   *     // a new callback or new options, and drops any existing destructor
+   *     // registrations so we avoid having extra destructors lying around.
+   *     disconnect(this);
+   *     unregisterDestructor(this, disconnect);
+   *
+   *     this.observer = new IntersectionObserver(callback, options);
+   *     this.observer.observe(element);
+   *
+   *     registerDestructor(this, disconnect);
```
rwjblue, line 122: Now that I read this again with fresh eyes, I think we _probably_ don't have to keep unregistering/reregistering since the actual thing we are doing **in this case** is using instance state to actually handle the disconnect (`instance.observer.disconnect()`) and the actual function is the same every time.

You would still need to either:

* unregister + reregister
* only register once

And which of those options that you choose probably depends on how your destructor works (e.g. does it use a stable instance prop, is it some other thing, etc).

---


FWIW, I think I'd personally do this for the case in the docs here (which only needs registration once, since the destructor is stable):

```js
function disconnect(instance) {
  instance.observer?.disconnect();
}

class IntersectionObserver extends Modifier {
  observer;

  constructor(owner, args) {
    super(owner, args);

    registerDestructor(this, disconnect);
  }

  modify(element, callback, options) {
    disconnect(this);
    this.observer = new IntersectionObserver(callback, options);
    this.observer.observe(element);
  }
}
```

chriskrycho: Ah, this seems better. Thank you!

rwjblue (approved):

PR #220, opened 2022-03-23
Deprecate didInstall, didReceiveArguments, and didUpdateArguments

chriskrycho:
As described in #217, all the functionality supplied by these hooks is now available, and with the desired "lazy"/standard autotracking semantics, providing a path for gradual migration. These APIs will be removed at 4.0.

### DIFF
--- a/addon/-private/class/modifier.ts
+++ b/addon/-private/class/modifier.ts
@@ -214,15 +214,19 @@ export default class ClassBasedModifier<S = DefaultSignature> {
    * @deprecated Until 4.0. Use `isDestroying` from `@ember/destroyables`.
    */
   get isDestroying(): boolean {
-    deprecate('Modifier.isDestroying is deprecated', false, {
-      id: 'ember-modifier.use-destroyables',
-      until: '4.0.0',
-      for: 'ember-modifier',
-      since: {
-        available: '3.2.0',
-        enabled: '3.2.0',
-      },
-    });
+    deprecate(
+      'Modifier.isDestroying is deprecated',
+      !SEEN_CLASSES_FOR_DESTROYABLES.has(this.constructor),
+      {
+        id: 'ember-modifier.use-destroyables',
+        until: '4.0.0',
+        for: 'ember-modifier',
+        since: {
+          available: '3.2.0',
+          enabled: '3.2.0',
+        },
+      }
+    );
 
     if (DEBUG && !SEEN_CLASSES_FOR_DESTROYABLES.has(this.constructor)) {
       SEEN_CLASSES_FOR_DESTROYABLES.add(this.constructor);
--- a/addon/-private/class/modifier.ts
+++ b/addon/-private/class/modifier.ts
@@ -16,7 +16,26 @@ function deprecateForDestroyables<S>(
   instance: ClassBasedModifier<S>
 ): void {
   deprecate(
-    `Modifier.${name} is deprecated`,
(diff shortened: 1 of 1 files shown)


addon/-private/class/modifier.ts
```diff
@@ -16,7 +16,26 @@ function deprecateForDestroyables<S>(
   instance: ClassBasedModifier<S>
 ): void {
   deprecate(
-    `Modifier.${name} is deprecated`,
+    `Modifier.${name} is deprecated. Use the corresponding API from '@ember/destroyable' instead.`,
```
rwjblue, line 19: Total bike shedding here, but when I see `Modifier.foo` I think you mean a static class method (not a prototype)

rwjblue (changes requested):
I think we also need some README documentation updates too

addon/-private/class/modifier.ts
rwjblue: Same basic issue as I mentioned in https://github.com/ember-modifier/ember-modifier/pull/221#discussion_r833596427, we need to include information about the instance so folks can actually figure out which modifiers are triggering the deprecation

rwjblue (changes requested):

addon/-private/class/modifier.ts
```diff
@@ -82,14 +76,51 @@ export default class ClassBasedModifier<S = DefaultSignature> {
...
+    );
+
+    if (DEBUG && !SEEN_CLASSES_FOR_DESTROYABLES.has(this.constructor)) {
+      SEEN_CLASSES_FOR_DESTROYABLES.add(this.constructor);
+    }
+
+    deprecate(
+      `ember-modifier (in ${this.constructor.name} at ${
+        new Error().stack
+      }): The \`didInstall\`, \`didReceiveArguments\`, and \`didUpdateArguments\` hooks are deprecated. Use the new \`modify\` hook instead.`,
```
rwjblue, line 107: Should we add error stack info to the `willDestroy`/`isDestroying`/`isDestroyed` ones too? Seems generally good, but if you do want to do that I would **not** try to share the stack; I can't think of a way to share that would avoid eagerly computing the `.stack` property (which is pretty expensive) even when you might not trigger a deprecation.

addon/-private/class/modifier.ts
chriskrycho: Fixed both of these issues (and it's good bike shedding!).

addon/-private/class/modifier.ts
chriskrycho: Yes, I think so, mean to do so!

addon/-private/class/modifier.ts
```diff
@@ -185,22 +224,34 @@ export default class ClassBasedModifier<S = DefaultSignature> {
       },
     });
 
+    if (DEBUG && !SEEN_CLASSES_FOR_DESTROYABLES.has(this.constructor)) {
```
rwjblue, line 227: The deprecation just above needs to change from `false` to leverage `SEEN_CLASSES_FOR_DESTROYABLES.has(this.constructor)`, right now it always emits a deprecation

rwjblue (changes requested):

addon/-private/class/modifier.ts
chriskrycho: Ahhh, thank you!

rwjblue (approved):

PR #221, opened 2022-03-23
Deprecate `this.element` and `this.args` on class-based modifiers

chriskrycho:
The introduction of the `modify` hook on class-based modifiers allows users to use the `element` and both `positional` and `named` arguments directly, as arguments to `modify`, rather than having them as the class properties `this.element` and `this.args` (see #217). Accordingly, deprecate access to both `this.element` and `this.args`.

(Users who want to stash the `element` on their modifier class can continue to do so directly, by assigning it from the `modify` call.)

### DIFF
--- a/addon/-private/class/modifier.ts
+++ b/addon/-private/class/modifier.ts
@@ -14,14 +14,25 @@ import { DEBUG } from '@glimmer/env';
 
 // SAFETY: these sets are dev-only code to avoid showing deprecations for the
 // same class more than once.
-let SEEN_CLASSES_FOR_LIFECYCLE: Set<ClassBasedModifier['constructor']>;
+type ClassBasedModifierClass = ClassBasedModifier['constructor'];
+let SEEN_CLASSES_FOR_LIFECYCLE: Set<ClassBasedModifierClass>;
 if (DEBUG) {
   SEEN_CLASSES_FOR_LIFECYCLE = new Set();
 }
 
-let SEEN_CLASSES_FOR_DESTROYABLES: Set<ClassBasedModifier['constructor']>;
+let SEEN_CLASSES_FOR_DESTROYABLES: Set<ClassBasedModifierClass>;
 if (DEBUG) {
   SEEN_CLASSES_FOR_DESTROYABLES = new Set();
+}
+
+let SEEN_CLASSES_FOR_ARGS: Set<ClassBasedModifierClass>;
+if (DEBUG) {
+  SEEN_CLASSES_FOR_ARGS = new Set();
+}
+
+let SEEN_CLASSES_FOR_ELEMENTS: Set<ClassBasedModifierClass>;
+if (DEBUG) {
+  SEEN_CLASSES_FOR_ELEMENTS = new Set();
 }
 
 /** @internal */
@@ -39,6 +50,12 @@ export const _implementsLegacyHooks = <S>(
   instance.didReceiveArguments !==
     ClassBasedModifier.prototype.didReceiveArguments;
 
+/** @internal */
+export const Element = Symbol('Element');
+
+/** @internal */
+export const Args = Symbol('Args');
+
 /**
  * A base class for modifiers which need more capabilities than function-based
  * modifiers. Useful if, for example:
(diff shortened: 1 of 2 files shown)


addon/-private/class/modifier.ts
```diff
@@ -60,27 +66,87 @@ export const _implementsLegacyHooks = <S>(
...
+
+    // SAFETY: the point here is (for the period where we are providing `args`
+    // and `element`) to provide an internal-only way of setting and update the
+    // `args` for the modifier instance; we use the `InternalClassBasedModifier`
+    // interface to represent the internal-only API in a way that end users do
+    // *not* have access to when subclassing `ClassBasedModifier`.
+    (this as unknown as InternalClassBasedModifier<S>)[Args] = args;
+
+    // TODO: remove both of these `defineProperty`s at v4.
+    Object.defineProperty(this, 'args', {
```
rwjblue, line 109: I'm curious, why not make it a getter in the class body (specifically, why use `Object.defineProperty` for **every** instance)?

I would expect this to work the same:

```js
class Modifier {
  get args() {
    deprecate(...);
    return this[Args];
  }
}
```

addon/-private/class/modifier.ts
```diff
@@ -60,27 +66,87 @@ export const _implementsLegacyHooks = <S>(
...
+            },
+            until: '4.0.0',
+          }
+        );
+
+        return this[Args];
+      },
+    });
+
+    Object.defineProperty(this, 'element', {
```
rwjblue, line 130: Same thing here, any reason not to use `get element() {}` instead of `Object.defineProperty` on every instance?

addon/-private/class/modifier.ts
```diff
@@ -60,27 +66,87 @@ export const _implementsLegacyHooks = <S>(
...
+    // interface to represent the internal-only API in a way that end users do
+    // *not* have access to when subclassing `ClassBasedModifier`.
+    (this as unknown as InternalClassBasedModifier<S>)[Args] = args;
+
+    // TODO: remove both of these `defineProperty`s at v4.
+    Object.defineProperty(this, 'args', {
+      enumerable: true,
+      get(this: InternalClassBasedModifier<S>) {
+        deprecate(
+          'ember-modifier: using `this.args` is deprecated. Access positional and named arguments directly in the `modify` hook instead.',
```
rwjblue, line 113: I think we should try to avoid issuing this deprecation when we are **also** issuing deprecations for lifecycle methods. By definition 100% of folks using the legacy hooks are gonna _also_ get this deprecation but that secondary / tertiary deprecation doesn't really provide any additional benefit (other than flooding console).

Also, these deprecations need to include **something** to help indicate _which_ modifier is triggering the deprecation. As it stands, you could get 3x deprecations from **every** modifier use in your app, with _no_ indication of which ones are wrong....

rwjblue (changes requested):

addon/-private/class/modifier.ts
chriskrycho: The comment on the field itself explains: we have to do this because existing subclasses used the fields as a place to define the types for the subclasses, and `field: Type` is a different type than `get field: Type`.

chriskrycho: That said, we *should* do this as `Object.defineProperty(Modifier.prototype, 'args', ...)` instead so we don't pay the perf hit!

addon/-private/class/modifier.ts
chriskrycho: Following up here after some conversations over Discord and also on #220: ultimately, we *do* need to deprecate these separately from each other and from the lifecycle hooks, because we don't have a type-safe path to *not* including them when we have `modify` only, and so folks *can* use `this.args` or `this.element` when in a `modify` hook, and *can* use only one or the other (though yes, that would be shenanigans).

chriskrycho:
For the sake of keeping merges clean and easy, this is now based on #220 and should land right after it does.

rwjblue (commented):
I think this needs a rebase now?

addon/-private/class/modifier.ts
```diff
@@ -261,4 +295,71 @@ export default class ClassBasedModifier<S = DefaultSignature> {
...
+// these, rather than one per instance. We also only issue the deprecation once
+// per class for each of `args` and `element`.
+Object.defineProperty(ClassBasedModifier.prototype, 'args', {
+  enumerable: true,
+  get(this: InternalClassBasedModifier<unknown>) {
+    deprecate(
+      `ember-modifier (in ${this.constructor.name} at ${
+        new Error().stack
+      }): using \`this.args\` is deprecated. Access positional and named arguments directly in the \`modify\` hook instead.`,
+      !SEEN_CLASSES_FOR_ARGS.has(this.constructor),
```
rwjblue, line 308: I think we should also avoid the deprecation for this case:

https://github.com/ember-modifier/ember-modifier/blob/d277c490a4b5ae3d03df29443095b8eeac5ae294/addon/-private/class/modifier.ts#L111

Specifically, we should try to avoid emitting a deprecation for needing to migrate to `.modify` **and** accessing `args`

addon/-private/class/modifier.ts
```diff
@@ -261,4 +295,71 @@ export default class ClassBasedModifier<S = DefaultSignature> {
...
+});
+
+Object.defineProperty(ClassBasedModifier.prototype, 'element', {
+  enumerable: true,
+  get(this: InternalClassBasedModifier<unknown>) {
+    deprecate(
+      `ember-modifier (in ${this.constructor.name} at ${
+        new Error().stack
+      }): using \`this.element\` is deprecated. Access the \`element\` as an argument in the \`modify\` hook instead.`,
+      !SEEN_CLASSES_FOR_ELEMENTS.has(this.constructor),
```
rwjblue, line 335: Same thing as above, I think if we've issued the lifecycle deprecation we shouldn't also issue this one

rwjblue (changes requested):

addon/-private/class/modifier.ts
rwjblue: chatted with @chriskrycho, this is still helpful for folks doing smaller transitions (e.g. rename `didReceiveArguments` to `modify` direclty), so they still see the other deprecated things to clean up...

rwjblue (approved):

PR #222, opened 2022-03-24
Introduce `eager` option to function-based modifiers

chriskrycho:
Introduce an options argument to the `modifier()` function used to create function-based modifiers, with a single initial option: `eager`. Passing `{ eager: false }` opts out of the previous default beahvior of eagerly consuming all arguments, so that function-based modifiers can have normal auto-tracking semantics -- that is, only rerunning when the tracked state they use changes. This allows users to opt into the new (`eager: false`) behavior on a per-modifier basis.

There are three changes to accompany the new options object here:

1.  Add overloads to `modifier` for the following cases:

    - the user passes no options (the default up till now)
    - the user passes `{ eager: true }` (matching the previous default)
    - the user passes `{ eager: false }` (the new target behavior)

    These overloads intentionally do *not* support invoking with a "signature" for `{ eager: false }`. If users want to introduce an explicit signature for function-based modifiers, they should first convert to the `{ eager: false }` semantics.

2.  Update the function-based modifier manager internals to support this. As a result, we now need to create a new instance each time, with the corresponding options.

3.  Update the internals of the function-based modifier manager to use a state bucket, aligning it with the internals of the class-based modifier manager and getting rid of the need for `Map`s for keeping track of modifier elements and teardowns.

---

(Docs and deprecation of the legacy behavior will follow!)

### DIFF
--- a/addon/-private/functional/modifier-manager.ts
+++ b/addon/-private/functional/modifier-manager.ts
@@ -1,98 +1,101 @@
 import { capabilities } from '@ember/modifier';
 import { gte } from 'ember-compatibility-helpers';
-import type { FunctionalModifierDefinition } from './modifier';
+import type { FunctionalModifierDefinition, Teardown } from './modifier';
 import type { ArgsFor, ElementFor } from '../signature';
 import { consumeArgs, Factory, isFactory } from '../compat';
-import { assert } from '@ember/debug';
 
-type ElementMap = {
-  get<S>(key: FunctionalModifierDefinition<S>): ElementFor<S> | null;
-  set<S>(key: FunctionalModifierDefinition<S>, value: ElementFor<S>): void;
-};
-
-type TeardownMap = {
-  get<S>(key: FunctionalModifierDefinition<S>): unknown;
-  set<S>(key: FunctionalModifierDefinition<S>, value: unknown): void;
-};
-
-const MODIFIER_ELEMENTS = new WeakMap() as ElementMap;
-const MODIFIER_TEARDOWNS = new WeakMap() as TeardownMap;
-
-function teardown<S>(modifier: FunctionalModifierDefinition<S>): void {
-  const teardown = MODIFIER_TEARDOWNS.get(modifier);
-
-  if (teardown && typeof teardown === 'function') {
-    teardown();
-  }
+interface State<S> {
+  instance: FunctionalModifierDefinition<S>;
 }
 
-function setup<S>(
-  modifier: FunctionalModifierDefinition<S>,
-  element: ElementFor<S>,
-  args: ArgsFor<S>
-): void {
-  const { positional, named } = args;
-  const teardown = modifier(element, positional, named);
-
-  MODIFIER_TEARDOWNS.set(modifier, teardown);
+interface CreatedState<S> extends State<S> {
+  element: null;
 }
 
-class FunctionalModifierManager<S> {
+interface InstalledState<S> extends State<S> {
+  element: ElementFor<S>;
+  teardown?: Teardown;
+}
+
+// Wraps the unsafe (b/c it mutates, rather than creating new state) code that
+// TS does not yet understand.
+function installElement<S>(
+  state: CreatedState<S>,
+  element: ElementFor<S>
+): InstalledState<S> {
+  // SAFETY: this cast represents how we are actually handling the state machine
+  // transition: from this point forward in the lifecycle of the modifier, it
+  // always behaves as `InstalledState<S>`. It is safe because, and *only*
+  // because, we immediately initialize `element`. (We cannot create a new state
+  // from the old one because the modifier manager API expects mutation of a
+  // single state bucket rather than updating it at hook calls.)
+  const installedState = state as State<S> as InstalledState<S>;
+  installedState.element = element;
+  return installedState;
+}
+
+export default class FunctionalModifierManager<S> {
   capabilities = capabilities(gte('3.22.0') ? '3.22' : '3.13');
+
+  options: { eager: boolean };
+
+  constructor(options?: { eager: boolean }) {
+    this.options = {
+      eager: options?.eager ?? true,
+    };
+  }
 
   createModifier(
     factoryOrClass:
       | Factory<FunctionalModifierDefinition<S>>
       | FunctionalModifierDefinition<S>
-  ): FunctionalModifierDefinition<S> {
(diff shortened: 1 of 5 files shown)


addon/-private/functional/modifier.ts
```diff
@@ -103,12 +163,12 @@ export default function modifier(
   // returns an opaque `Modifier` type so that we can provide a result from this
   // type which is useful to TS-aware tooling (e.g. Glint).
   return setModifierManager(
-    () => FunctionalModifierManager,
+    () => new FunctionalModifierManager(options),
```
rwjblue, line 166: Instead of making a new instance per `modifier(` call, let's just make two: one for eager and one for lazy; then hand back those singeltons...

rwjblue (changes requested):

addon/-private/functional/modifier.ts
chriskrycho: Ahhh, thank you! This was the one thing I was thinking about.

addon/-private/functional/modifier-manager.ts
```diff
@@ -1,98 +1,101 @@
...
+  }
+
   createModifier(
     factoryOrClass:
       | Factory<FunctionalModifierDefinition<S>>
       | FunctionalModifierDefinition<S>
-  ): FunctionalModifierDefinition<S> {
-    const Modifier = isFactory(factoryOrClass)
+  ): CreatedState<S> {
+    const instance = isFactory(factoryOrClass)
```
rwjblue, line 53: minor note here: this ternary should be removed in 4.0 (when we drop < ember-source@3.22 support)

addon/-private/functional/modifier-manager.ts
```diff
@@ -1,98 +1,101 @@
...
-
-    setup(modifier, element, args);
   }
 
-  destroyModifier(modifier: FunctionalModifierDefinition<S>): void {
-    teardown(modifier);
+  destroyModifier(state: InstalledState<S>): void {
+    if (state.teardown) {
+      state.teardown();
+    }
```
rwjblue, line 99: Pretty syntax?
```suggestion
    state.teardown?()
```

addon/-private/functional/modifier.ts
```diff
@@ -54,14 +60,72 @@ type Teardown = () => unknown;
 export default function modifier<
   E extends Element,
   P extends unknown[],
-  N extends Record<string, unknown>
+  N extends object
```
rwjblue, line 63: I don't fully grok this change, what does changing to `object` vs `Record<string, unknown>` get us?

addon/-private/functional/modifier-manager.ts
chriskrycho: Yep, noted the same.

addon/-private/functional/modifier-manager.ts
chriskrycho: I personally hate reading `state.teardown?.()` 😂

addon/-private/functional/modifier.ts
```diff
@@ -54,14 +60,72 @@ type Teardown = () => unknown;
...
   Element: E;
 }>;
 
+/**
+ * An API for writing simple modifiers.
+ *
+ * This function runs the first time when the element the modifier was applied
+ * to is inserted into the DOM, and it *autotracks* while running. Any values
+ * that it accesses will be tracked, including the arguments it receives, and if
+ * any of them changes, the function will run again.
```
rwjblue, line 77: Since this is the docs for `eager: true` side, should this also mention that all of the incoming arguments are also eagerly consumed and will force the function to run again?

addon/-private/functional/modifier.ts
```diff
@@ -54,14 +60,72 @@ type Teardown = () => unknown;
...
+  Element: E;
+}>;
+
+/**
+ * An API for writing simple modifiers.
+ *
+ * This function runs the first time when the element the modifier was applied
+ * to is inserted into the DOM, and it *autotracks* while running. Any values
+ * that it accesses will be tracked, including the arguments it receives, and if
+ * any of them changes, the function will run again.
```
rwjblue, line 106: Maybe reword slightly to ensure that it's clear that _only_ arguments that are "consumed" are included in the "will run again" part.

addon/-private/functional/modifier.ts
```diff
@@ -89,26 +154,28 @@ export default function modifier(
   fn: (
     element: Element,
     positional: unknown[],
-    named: Record<string, undefined>
-  ) => void | Teardown
+    named: object
+  ) => void | Teardown,
+  options?: { eager: boolean }
```
rwjblue, line 159: Can we default here?
```suggestion
  options?: { eager: boolean } = { eager: true }
```

This makes it a bit clearer what the current default is (to my eyes at least)

addon/-private/functional/modifier.ts
```diff
@@ -89,26 +154,28 @@ export default function modifier(
...
-    Named: Record<string, unknown>;
+    Named: object;
     Positional: unknown[];
   };
 }> {
   // SAFETY: the cast here is a *lie*, but it is a useful one. The actual return
   // type of `setModifierManager` today is `void`; we pretend it actually
   // returns an opaque `Modifier` type so that we can provide a result from this
   // type which is useful to TS-aware tooling (e.g. Glint).
+  const isEager = !options || options?.eager;
```
rwjblue, line 171: This const should be moved _up_ above the `SAFETY` note just above; the note is talking about the line just below the `const` (the `return`) and having hte const between them confused me initially 😛

addon/-private/functional/modifier-manager.ts
rwjblue: heheheehe fair enough!

addon/-private/functional/modifier.ts
chriskrycho: Good call, yes.

addon/-private/functional/modifier.ts
chriskrycho: Ahhh, heh, whoops, yes. Thank you!

addon/-private/functional/modifier.ts
chriskrycho: Ah! Also lets us simplify the body a bit, too: it can just become a check on `options.eager` directly.

rwjblue (approved):

PR #223, opened 2022-03-24
Deprecate `modifier()` without `{ eager: false }`

chriskrycho:
This assumes, and should land immediately after, #222.

### DIFF
--- a/addon/-private/functional/modifier.ts
+++ b/addon/-private/functional/modifier.ts
@@ -1,3 +1,4 @@
+import { deprecate } from '@ember/debug';
 import { setModifierManager } from '@ember/modifier';
 import {
   DefaultSignature,
@@ -57,6 +58,11 @@ export type Teardown = () => unknown;
  * being removed entirely. It should generally clean up the changes that the
  * modifier made in the first place.
  *
+ * @deprecated Until 4.0. Calling `modifier()` without an options argument is
+ *   deprecated. It is supported until 4.0 so that existing modifiers can be
+ *   migrated individually. Please update your function-based modifiers to pass
+ *   `{ eager: false }`.
+ *
  * @param fn The function which defines the modifier.
  */
 // This overload allows users to write types directly on the callback passed to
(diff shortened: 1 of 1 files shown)


addon/-private/functional/modifier.ts
```diff
@@ -13,6 +15,15 @@ import FunctionalModifierManager from './modifier-manager';
...
 const LAZY_MANAGER = new FunctionalModifierManager({ eager: false });
 
+// Dev-only util to make sure we only log the options deprecation once per
+// modifier definition. Disabling the lint because in this case the point is
+// that we don't care about what the internal signature is.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let SEEN_FNS: Set<FunctionalModifierDefinition<any>>;
+if (DEBUG) {
+  SEEN_FNS = new Set();
+}
```
rwjblue, line 25: I actually think we can get away without doing this trick here. Specifically because modules themselves already cache, and we are emitting all of the new deprecations during the initial module evaluation (because that is when `modifier(...)` is actually called).

In other words, I'm pretty sure that even if you remove this set we cannot have multiple deprecations for the same modifier.

rwjblue (changes requested):
Overall, LGTM. One minor inline comment, and I think it needs a rebase/merge/etc to drop out the commits from #222.

addon/-private/functional/modifier.ts
chriskrycho: Ah! I hadn’t thought of that—thank you!

beerinho:
Hey @chriskrycho, what is the plan to update the docs with the changes in this PR? I updated ember-modifier yesterday to 3.2.0 and am getting the deprecation about not passing options, I feel like I have updated my modifier to pass in `{eager: false}` but I still get the deprecation warning and I couldn't find anything in the docs to help me figure out where I'm going wrong. My modifier looks like this:
```
  return modifier(
    function suggestModifier(element, [function_, ...args], named) {
        ...function code
    },
    { eager: false }
  );
```

chriskrycho:
The docs *are* updated! Additionally there is an entire migration guide. That said, I’m sorry you’re having this issue; please open an issue with a more detailed report so we can help you get to the bottom of it (and possibly fix the deprecation if it’s firing when it shouldn’t be).

chriskrycho:
I believe this is probably #249. I'll have a fix out for it in a while.

PR #227, opened 2022-03-25
Update docs and export type utils

chriskrycho:
- Update all docs to cover new `modify()` API and its normal auto-tracking semantics.
- Add a migration guide for v4.
- Export a handful of type signatures that will be useful for consumers working explicitly with the new `Signature` type.

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -37,7 +37,7 @@ The point is to give you a path to decouple the upgrade and the migration work:
 
 ### Function-based modifiers
 
-For function-based modifiers, you simply need to pass a new options argument to the `modifier()` call.
+For function-based modifiers, the only change you need to make is to pass a new options argument to the `modifier()` call.
 
 Previously, you would write this:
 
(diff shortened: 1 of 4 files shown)


MIGRATIONS.md
```diff
@@ -0,0 +1,356 @@
...
+### Class-based modifiers
+
+For class-based modifiers,  you need to migrate to the new `modify()` API. *All* of the old lifecycle hooks are deprecated, as are the `element`, `args`, and `isDestroying` and `isDestroyed` fields. Additionally, the  new `modify()` hook which replaces the previous lifecycle hooks is *lazy*, like auto-tracking in general. After installation, it will only be re-executed when some tracked state it uses actually changes, where previously it.
+
+
+#### Lifecycle hooks
+
+##### `didInstall()`
+
+For one-time setup done with the element in `didInstall()`, do the same setup in `modify()`. If it is cheap, you can just let it happen each time. If it is expensive, or if the operation is not [idempotent](https://en.wikipedia.org/wiki/Idempotence), you can set a flag to avoid doing it again:
```
rwjblue, line 81: Might be nice to have an example in the normal docs (not these migration docs though probably) of how you can do this same kinda "flag" deal but still have it be dependent on some incoming state by using tracked cache (e.g. `import { createStorage, getValue, setValue } from 'ember-tracked-storage-polyfill'`).

PR #245, opened 2022-03-30
Update Class-based modifier example

patocallaghan:
This example references `this.removeEventListener` which doesn't exist. I've replaced it with the `cleanup` function which exists at the top of the example. Or should the `cleanup` function be refactored to be a `removeEventListener` method on the class instead?


chriskrycho (approved):
Ah, thank you for this! This is correct – using the standalone function is generally best practice where possible with `@ember/destroyable`, so that it doesn't have to be bound and doesn't have to be per-instance.

PR #247, opened 2022-03-30
Fix "default" positional and named argument types

chriskrycho:
Previously, we resolved the default types of positional and named arguments to `unknown` and `Record<string, unknown>` respectively in TS context. However, Glint needs those to be `[]` (an empty tuple) and a special `EmptyObject` type respectively, so that it can show errors when users pass positional parameters to a type where.

This *only* applies to new `Signature` types. It does not change the the previous defaults of `unknown[]` and `Record<string, unknown>` when leaving the types unspecified entirely, and does not affect inference when using non-`Signature` type params explicitly.

Note: this default *will* be changed in v4.0 so *all* unspecified named and positional arguments get the new defaults, so they work correctly with Glint.


package.json
```diff
@@ -89,7 +89,7 @@
     "release-it": "^14.11.6",
     "release-it-lerna-changelog": "^4.0.1",
     "rimraf": "^3.0.2",
-    "typescript": "^4.5.2",
+    "typescript": "^4.7.0-dev.20220330",
```
dfreeman, line 92: Did you want to keep this change?

dfreeman (approved):

package.json
chriskrycho: Ugh no

PR #253, opened 2022-03-31
Check type of teardown before saving or using it

### LINKED ISSUE #250: [3.2.0]+ `state.teardown is not a function`
After updating from 3.1.0 to 3.2.1, I get a runtime exception:

```
Uncaught (in promise) TypeError: state.teardown is not a function
    at FunctionBasedModifierManager.destroyModifier (modifier-manager.js:75:1)
```

My modifier is pretty simple:

```ts
import { set } from '@ember/object';
import { modifier } from '@glint/environment-ember-loose/ember-modifier';
import type { ExtractFieldsOfType } from '...';

const refModifier = modifier(<T, U extends ExtractFieldsOfType<T, HTMLElement>>(element: Element & T[U], [context, property]: [T, U]) =>
  set(context, property, element),
);

export default refModifier;
```

The `state` variable in `destroyModifier` is:

```
element: a#ember322.ember-view
instance: (element, _ref) => {…}
teardown: a#ember322.ember-view
[[Prototype]]: Object
```

I.e. it has `teardown` but it's definitely not a function but rather some element. Any ideas?

chriskrycho:
Previously, this path "trusted" callers to only return functions in function-based modifier definitions, in line with the requirement that teardown can only be a function. However, one reasonably-common pattern with modifiers is to simply write an arrow function which performs the operation, and whose expression value is therefore the return value of the modifier definition.

For example, it's possible (though very much not advisable!) to do something like this to set a reference to the element on the passed-in object:

```js
import { set } from '@ember/object';
import { modifier } from 'ember-modifier';

export default modifier((el, [obj]) => set(obj, 'element', el));
```

This will return the *element*. We could `assert` on this, but that would make an otherwise unobjectionable pattern less friendly (even if this particular example is not one we want to encourage).

Accordingly, do not save or attempt to run non-`function` types as teardowns in function-based modifiers.

Fixes #250

### DIFF
--- a/addon/-private/function-based/modifier-manager.ts
+++ b/addon/-private/function-based/modifier-manager.ts
@@ -69,7 +69,7 @@ export default class FunctionBasedModifierManager<S> {
 
     const { positional, named } = args;
     const teardown = createdState.instance(element, positional, named);
-    if (teardown) {
+    if (typeof teardown === 'function') {
       state.teardown = teardown;
     }
 
@@ -84,7 +84,7 @@ export default class FunctionBasedModifierManager<S> {
     }
 
     const teardown = state.instance(state.element, args.positional, args.named);
-    if (teardown) {
+    if (typeof teardown === 'function') {
       state.teardown = teardown;
     }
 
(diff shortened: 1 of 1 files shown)


boris-petrov:
Thanks for fixing this! Why is it `very much not advisable` if it works? :)

PR #270, opened 2022-04-04
Use "export type" for FunctionBasedModifier interface

ef4:
Without this change you get a broken reexport when built by babel, because it doesn't know to strip `FunctionBasedModifier` out of index.ts, but it does know to strip it out of -private/function-based/modifier.ts.

This results in a warning from webpack when using embroider. It's also technically a SyntaxError to try to reexport a thing that doesn't exist.

### DIFF
--- a/addon/index.ts
+++ b/addon/index.ts
@@ -1,6 +1,4 @@
 export { default } from './-private/class/modifier';
-export {
-  default as modifier,
-  FunctionBasedModifier,
-} from './-private/function-based/modifier';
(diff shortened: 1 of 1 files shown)


ef4:
(I opened this PR through the GitHub web editor and wasn't very familiar with its defaults!)

chriskrycho:
Thanks for this! I'll land it tomorrow (and back port to v3 as well).

chriskrycho:
Update: it's out now as v3.2.4.

PR #277, opened 2022-04-05
Make `FunctionBasedModifier` an (abstract) constructor type

dfreeman:
Similar to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59714, this PR makes `FunctionBasedModifier` an abstract constructor type. This change allows `modifier()` to accept generic functions, preserving their type parameter(s) in the type of the resulting value by virtue of it being a bare constructor type.

This type is declared to produce an abstract subtype of the class-based `Modifier`, so that a single augmentation from a tool like Glint will apply to all modifiers produced from this package, whether function- or class-based.

### DIFF
--- a/addon/-private/function-based/modifier.ts
+++ b/addon/-private/function-based/modifier.ts
@@ -1,22 +1,36 @@
 import { setModifierManager } from '@ember/modifier';
-import Opaque from '../opaque';
 import {
   ElementFor,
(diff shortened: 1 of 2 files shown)


chriskrycho (approved):
Thank you! I will back-port to v3 and release!

PR #315, opened 2022-04-20
Update class-based modifier in blueprint

SergeAstapov:

PR #316, opened 2022-04-20
Update function-based modifier in blueprint

SergeAstapov:
This is related to #315.

This would allow to generate new function-based modifier without deprecation in v3.
This is not needed in v4 as only `true` value supported for `eager` option.


chriskrycho (approved):
Excellent fix. Thank you!

PR #326, opened 2022-04-22
Fix `watchDependencies` usage

SergeAstapov:
This was incorrectly setup during convertion to v2 addon format in #296


chriskrycho (commented):
👍🏼

PR #327, opened 2022-04-22
auto-publish unstable packages to NPM

SergeAstapov:
As there's no convenient way to point NPM directly at a monorepo, and even if you could the package does have a required build step.

The build step for v2 addons was a deliberate design choice, because in the normal case apps can have much faster builds if all their addons are already prebuilt before publishing. But it does make the case of using unreleased code more manual.

This PR adds workflow to automatically publish alpha releases to NPM for every commit to the master branch.

This is inspired by setup in https://github.com/ember-animation/ember-animated/blob/master/.github/workflows/publish-unstable.yml

@chriskrycho would you be open doing this? If yes - could you please add npm token as `NODE_AUTH_TOKEN` secret to the repository config?

### DIFF
--- a/.github/workflows/publish-unstable.yml
+++ b/.github/workflows/publish-unstable.yml
@@ -1,0 +1,41 @@
+# For every push to the master branch, this publishes an NPM package to the
+# "unstable" NPM tag.
+
+name: Publish Unstable
+
+on:
+  push:
+    branches:
+      - main
+      - master
+
+jobs:
+  publish:
+    name: "NPM Publish"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14.x
+          cache: yarn
+          # This creates an .npmrc that reads the NODE_AUTH_TOKEN environment variable
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install Dependencies
+        run: yarn install --frozen-lockfile
+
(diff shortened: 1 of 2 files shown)


chriskrycho:
Seems good – remind me and I'll get it set up next week!

SergeAstapov:
```
npm notice === Tarball Details === 
npm notice name:          ember-modifier                                  
npm notice version:       4.0.0-beta.1-unstable.2f19c32                   
npm notice filename:      ember-modifier-4.0.0-beta.1-unstable.2f19c32.tgz
npm notice package size:  16.1 kB                                         
npm notice unpacked size: 53.5 kB
```

woohoo! it got automatically published with `unstable` tag

chriskrycho:
👏🏼 👍🏼 🎉 🚀 Thanks for getting that taken care of!

Turbo87:
keep in mind that something like `4.0.0-unstable.f00` is higher in semver order than `4.0.0-beta.1` because prerelease tags are ordered alphabetically

SergeAstapov:
@Turbo87 It gets published with `unstable` tag so should not interfere

chriskrycho:
The tag is distinct from the ordering. If someone asks for `@unstable` they’ll get this; but if someone asks for `^4.0.0-beta.1` they will *also* get it.

SergeAstapov:
thanks for explanation @chriskrycho @Turbo87
Do you think we should roll this back or skip pre-release versions?

chriskrycho:
I think pre-release versions are fine, but let's just mark them as `-alpha` instead.

PR #329, opened 2022-04-22
The package must be public

chriskrycho:
When converting to a monorepo, we accidentally made the addon package itself `"private": true` as well as the test app and the root package. This resulted in our not actually publishing it!


SergeAstapov:
@chriskrycho oh no! I'm sorry about that mistake. I only can image how trolling it was

chriskrycho:
Haha we’ve all been there!

PR #331, opened 2022-04-23
Remove `-private/signature.ts` entrypoint

SergeAstapov:
This was identified in https://github.com/ember-modifier/ember-modifier/pull/296#discussion_r854579961

### DIFF
--- a/addon/src/index.ts
+++ b/addon/src/index.ts
@@ -1,4 +1,9 @@
 export { default } from './-private/class-based/modifier';
 export { default as modifier } from './-private/function-based/modifier';
 export type { FunctionBasedModifier } from './-private/function-based/modifier';
-export type { ArgsFor, NamedArgs, PositionalArgs } from './-private/signature';
+export type {
+  ArgsFor,
+  DefaultSignature,
+  NamedArgs,
+  PositionalArgs,
+} from './-private/signature';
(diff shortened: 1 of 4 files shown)


addon/src/index.ts
```diff
@@ -1,4 +1,9 @@
 export { default } from './-private/class-based/modifier';
 export { default as modifier } from './-private/function-based/modifier';
 export type { FunctionBasedModifier } from './-private/function-based/modifier';
-export type { ArgsFor, NamedArgs, PositionalArgs } from './-private/signature';
+export type {
+  ArgsFor,
+  DefaultSignature,
```
chriskrycho, line 6: Per https://github.com/ember-modifier/ember-modifier/pull/296#discussion_r854654276, what we actually want to do here is a *little* different: we want to *not* export `DefaultSignature` this way, but instead, using [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) and [self-referencing](https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name), we want to make it so we *can* import these things from ourselves locally… but no one else can.

Net, I believe that should look like this:

`package.json`:

```js
{
  "imports": {
    "#ember-modifier": {
      /* mapping just like in `exports` */
    }
  }
}
```

Minimally, if this doesn't work for some reason, we can just use relative imports instead. 😅

PR #332, opened 2022-04-23
enable `embroider-optimized` scenario

SergeAstapov:


SergeAstapov:
> @SergeAstapov do you know why this is failing?

it complains about dynamic modifier:

```
 $TMPDIR/embroider/cae3d9/test-app/tests/integration/modifiers/class-modifier-test.ts/class-modifier-test.ts: Missing modifier: using-constructor in tests/integration/modifiers/class-modifier-test.ts
```

basically this line https://github.com/ember-modifier/ember-modifier/blob/1c6c52d7289155049bcbcaf1d03b19a6523161bf/test-app/tests/integration/modifiers/class-modifier-test.ts#L44

the safe way of making Embroider happy is to do instead

```
this['using-constructor'] = UsingConstructor;

await render(hbs`
  <h1 id="expected" {{this.using-constructor this.state.pos named=this.state.named}}>Hello</h1>
`);
```

BUT this does not work in Ember 3.24 which is being tested by `ember-try`.

So I may need to find a way to make it simple yet still working in both worlds.

ef4:
Ember 3.24 LTS is past end of support. Starting at 3.25 first-class modifiers work so this problem goes away.

But if you want to make this test work in 3.24, you can put the modifiers in `test-app/app/modifiers` so that Embroider can resolve them. If they need to share state with the test context that can be done via a service.

chriskrycho:
Thanks! @SergeAstapov lets just drop support for 3.24. (Let's do that in a separate PR to make the change log clearer, since that is its own distinct breaking change.)

PR #450, opened 2022-07-20
Options for modifier deprecation warning

### LINKED ISSUE #449: Be more explicit in v3 -> v4 modifier() deprecation warning
# Problem

When calling `modifier(() => {})` I get following deprecation warning:

```
Until 4.0. Calling modifier() without an options argument is deprecated. It is supported until 4.0 so that existing modifiers can be migrated individually. Please update your function-based modifiers to pass { eager: false }.
```

Which is nice, but:

1. The message itself does not hint me *where* to put the `options` argument.
2. The README does not show that there is any `options` argument at all.

# Solution

1. Add specific example to the deprecation warning.
2. Make a note in readme that `modifier()` has `options` argument so that people can refer to it.

MichalBryxi:
1. Add specific example to the deprecation warning.
2. Make a note in readme that modifier() has options argument so that people can refer to it.
3. Fixes #449

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ Function-based modifiers consist of a function that receives:
 3. An object of named arguments
 
 ```js
-modifier((element, positional, named) => { /* */ });
+modifier((element, positional, named) => { /* */ }, options);
 ```
 
 This function runs the first time when the element the modifier was applied to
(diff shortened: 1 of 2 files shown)


chriskrycho (approved):
Thank you! Sorry about the delayed response!

PR #486, opened 2022-09-21
Update README examples that used deprecated behavior

DingoEatingFuzz:
In preparation for v4, function-based modifiers must now provide `{ eager: false }` as a second argument. Having code snippets in the README that rely on deprecated behavior makes for a bummer of an experience.

### DIFF
--- a/addon/README.md
+++ b/addon/README.md
@@ -173,7 +173,7 @@ export default modifier((element, [eventName, handler]) => {
   return () => {
     element.removeEventListener(eventName, handler);
   }
-});
+}, { eager: false });
 ```
 
 Here, we setup the event listener using the positional parameters passed to the
@@ -221,7 +221,7 @@ export default modifier((element, [eventName, handler]) => {
   return () => {
     element.removeEventListener(eventName, handler);
   }
-});
+}, { eager: false });
 ```
 
 Function-based modifiers consist of a function that receives:
@@ -298,7 +298,7 @@ export default modifier(element => {
   }, 1000);
 
   return () => clearInterval(id);
-});
+}, { eager: false });
 
 ```
 ```hbs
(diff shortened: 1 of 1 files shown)


SergeAstapov:
@DingoEatingFuzz I understand the confusion, but readme currently says:

> NOTE: this is the README for the v4 release (currently available in beta)! For the v3 README, see [here](https://github.com/ember-modifier/ember-modifier/blob/v3/README.md).

not sure tbh what should be ideal solution until v4 is out

DingoEatingFuzz:
Ah, so it does. I guess have a bad habit of scrolling right past the table of contents 😬 

But also, the v3 readme uses the deprecated function-based modifier pattern.

This is just my opinion, but I think if a project doesn't have a docs site, then the readme should represent the current released version. 

Like maybe have a README_V4.md and then cut it over to the main README.md when V4 is out of beta?

Or maybe it's not a big deal, just seemed like an easy itch to scratch last night.

SergeAstapov:
@DingoEatingFuzz I totally agree this may be frustrating.

@chriskrycho may have some ideas and if/when v4 can be released

chriskrycho:
@SergeAstapov related question: what's our status on making those couple tweaks to emberjs/rfcs#811? The only reason I’ve really been waiting for the v4 release here is because I want to make sure we have everything we need *before* releasing v4 so that the v4 release is the one we include with that!

@DingoEatingFuzz good callout—I originally expected us to have cut over months ago, and if it takes us substantially longer, we should do something similar to that.

SergeAstapov:
@chriskrycho oh I'm sorry, didn't have time for OSS due to personal reasons an then it felt off my radar, just pushed an update addressing the question

chriskrycho:
Totally understand—been there myself many a time!

johanrd:
@SergeAstapov this can be closed as not relevant, anymore, I guess.

PR #498, opened 2022-10-11
Update signature of constructor for class-based modifiers

SergeAstapov:


SergeAstapov:
this was a rabbit hole but I think figured out the path forward:

at first, CI was broken (on `Floating Dependencies` job) as we depend TS with caret - `"typescript": "^4.6.4"`, hence it tried to use the latest v4.8.4 which fails with `rollup-plugin-ts` v2.

Bumping `rollup-plugin-ts` to v3 resolved that issue but then I somehow touched yarn too many times and tried some other deps in hope to fix issues.
and as a result, started having issues with `setOwner` due to `owner: unknown` which was changed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61581

So this PR also had to update signature of constructor for class-based modifiers as well as tests and type-tests.

addon/src/-private/class-based/modifier-manager.ts
```diff
@@ -57,7 +58,7 @@ export default class ClassBasedModifierManager<S> {
     dependencySatisfies('ember-source', '^3.22.0 || ^4.0.0') ? '3.22' : '3.13'
   );
 
-  constructor(private owner: unknown) {}
+  constructor(private owner: Owner) {}
```
SergeAstapov, line 61: @chriskrycho do I get it right we should change the signature to adhere https://rfcs.emberjs.com/id/0821-public-types/#owner ?

as we are in v4.beta, this might be a good timing to do so as it's breaking change per types policy.
unless I'm missing something

addon/type-tests/modifier-test.ts
```diff
@@ -264,17 +265,14 @@ class ClassBased extends Modifier<ClassBasedSignature> {
...
 }
 
-const classBased = new ClassBased(
-  { iAmAnOwner: 'yep' },
-  {
-    named: {
-      onMessage(desc, data) {
-        console.log(desc, JSON.stringify(data, null, 2));
-      },
+const classBased = new ClassBased({ iAmAnOwner: 'yep' } as unknown as Owner, {
```
SergeAstapov, line 268: (sorry for indentation, it was not me but prettier) as we don't have real owner here, this should be ok, I think

addon/src/-private/class-based/modifier-manager.ts
chriskrycho: Yep, this is the right thing to do (and sorry for the long-delayed response)!

addon/type-tests/modifier-test.ts
chriskrycho: Seems fine, yeah; that's an annoying thing for tests. We *should* also be able to use the owner from the test context if it's useful, but there's a known gap in the need to be able to get an owner otherwise.

PR #599, opened 2022-12-13
Remove mention of v4 beta from readme

sandstrom:


SergeAstapov:
Thank you @sandstrom!

PR #600, opened 2022-12-13
Skip publish-unstable for dependabot PRs

SergeAstapov:


SergeAstapov:
this would reduce number of versions we publish to only those that have some meaningful changes.

chriskrycho (approved):
Makes sense to me! Good call.

PR #657, opened 2023-01-31
Include @ember/owner in tsconfig types

gitKrystan:
Without this config, projects using ember-source [preview types](https://blog.emberjs.com/announcing-official-typescript-types-public-preview/) see this TS error:

```
node_modules/ember-modifier/dist/index.d.ts:1:23 - error TS2688: Cannot find type definition file for 'ember__owner'.

1 /// <reference types="ember__owner" />
                        ~~~~~~~~~~~~
```

While this project doesn't add the reference directive directly, it appears that with `types: []` ([the default for @tsconfig/ember](https://github.com/tsconfig/bases/pull/149)), the compiler will helpfully(?) add the directive for you. (See https://github.com/microsoft/TypeScript/issues/19874)

With this config, type-checking in my dogfood project appears to work both with and without preview types.

### DIFF
--- a/addon/tsconfig.json
+++ b/addon/tsconfig.json
@@ -2,5 +2,8 @@
   "extends": "@tsconfig/ember/tsconfig.json",
   "include": [
     "src/**/*"
-  ]
+  ],
+  "compilerOptions": {
+    "types": ["@ember/owner"]
+  },
 }


gitKrystan:
I talked to @chriskrycho and @dfreeman about this [on the Discord](https://discord.com/channels/480462759797063690/786312479620726804/1070347293737369681). Here were there thoughts:

> dfreeman — 02/01/2023 6:18 AM
> The emit behavior is surprising to me, but given that that's apparently how it works, your fix seems reasonable to me 👍
> Slightly longer-term, the answer may be to move off of the `@types` packages there (maybe just with a test scenario that uses them instead of the native types) so that this isn't a hazard when publishing
> (Chris should confirm when he has the chance to look though)

> Chris Krycho — 02/01/2023 7:51 AM
> UGHGHGHHHHH
> I would definitely prefer Dan's second suggestion if you're up for it, @gitKrystan:
> move off of the `@types` packages there (maybe just with a test scenario that uses them instead of the native types) so that this isn't a hazard when publishing

I am happy to look into the longer term fix soon.

SergeAstapov:
@gitKrystan thank you for working on this! I assume Chris meant to do the same changes as were done in https://github.com/tracked-tools/ember-async-data/pull/473

@gitKrystan feel free to open such PR.

SergeAstapov:
@gitKrystan sidenote - please refresh branch with latest `master` to get green CI

gitKrystan:
I opened https://github.com/ember-modifier/ember-modifier/pull/665 with the preferred fix. I suspect that one requires a new major version though.

dfreeman:
Just weighing in with my own take here: I think it makes sense to move forward with this fix as a starting point to unblock folks using the native types.

We should be able to land #665 without requiring a new major, but part of doing that will be adding a test scenario that _proves_ everything still typechecks when you're using `@types` packages instead of the native types.

gitKrystan:
> We should be able to land https://github.com/ember-modifier/ember-modifier/pull/665 without requiring a new major, but part of doing that will be adding a test scenario that proves everything still typechecks when you're using @types packages instead of the native types.

@dfreeman Great idea. Do you have an example of that? I'm not sure how to go about it.

dfreeman:
Glint's [`test-packages`](https://github.com/typed-ember/glint/tree/main/test-packages) have two separate apps (`ts-ember-app` and `ts-ember-preview-types`) that exercise both, but I think it should be possible in simpler projects like this to just test both configurations against a single test app.

The issue is just that I don't know of anyone who's done that yet, so it may take a bit of experimentation—that's why I suggested shipping this fix in the meantime 🙂 

I'll think a bit on how the testing setup might look and comment on #665 if I come up with anything useful.

SergeAstapov:
@gitKrystan I assume we can close this PR?

gitKrystan:
Superseded by https://github.com/ember-modifier/ember-modifier/pull/665

PR #662, opened 2023-02-02
Use tilde instead of caret for rollup-plugin-ts

SergeAstapov:
newer rollup-plugin-ts pulls in a dependency that breaks on our node version support


SergeAstapov:
we may need to remove `engines.node` field from package.json of the addon as there is nothing addon has for node.js version support and then we could easily bump version of Node.js in CI

PR #665, opened 2023-02-06
Use the types published from Ember itself 

gitKrystan:
* Use Ember preview types.
* Drop the corresponding `@types` packages.

NOTE: This currently includes (and undoes) the commits from https://github.com/ember-modifier/ember-modifier/pull/657 as well. If we decide not to merge that one I can rebase this on master.

### Context

Without this change, projects using ember-source [preview types](https://blog.emberjs.com/announcing-official-typescript-types-public-preview/) see this TS error:

```
node_modules/ember-modifier/dist/index.d.ts:1:23 - error TS2688: Cannot find type definition file for 'ember__owner'.

1 /// <reference types="ember__owner" />
                        ~~~~~~~~~~~~
```

While this project doesn't add the reference directive directly, it appears that with `types: []` ([the default for @tsconfig/ember](https://github.com/tsconfig/bases/pull/149)), the compiler will helpfully(?) add the directive for you. (See https://github.com/microsoft/TypeScript/issues/19874)

With this change, type-checking in my dogfood project appears to work both with and without preview types.


addon/types/index.d.ts
```diff
@@ -0,0 +1,2 @@
+import 'ember-source/types';
```
NullVoxPopuli, line 1: in the v2-addon blueprint, we've been using a folder with more words 😅 
https://github.com/embroider-build/addon-blueprint/tree/main/files/__addonLocation__/unpublished-development-types

gitKrystan: Fixed

SergeAstapov: we should update addon to sync with latest updates in addon-blueprint, I hope I can do it sooner than later.

SergeAstapov:
@gitKrystan we could add a new test app with Ember native types in addition to the `test-app`.

however my concern is not about native types, but about bumping up peer dependency, which would screw up dependency graph, as addon may either get wrong `ember-source` or consuming app gets a problem from package manager if app uses `ember-source` lower than 4.8

gitKrystan:
> my concern is not about native types, but about bumping up peer dependency, which would screw up dependency graph, as addon may either get wrong ember-source or consuming app gets a problem from package manager if app uses ember-source lower than 4.8

Yeah, that's why I was thinking it would require a major version. I'm not really sure what the solution is, TBH. Hopefully someone with more experience in this area will have suggestions.

dfreeman:
You should be able to leave the peer dependency alone. The thing about the way `ember-source` publishing types works (since the actual `@ember/*` packages don't exist) is that it's effectively the host app's job to ensure that the declarations for `@ember/*` modules are ambiently available, whether that's via `import 'ember-source/types'` or the presence of `@types/ember*` packages.

dfreeman:
> we could add a new test app with Ember native types in addition to the `test-app`.

That's definitely one option! I linked to where we do that in the Glint monorepo in my comment on #657 

The other alternative might be to essentially manage it via `ember-try` and have the scenarios that pull in older Ember versions _also_ install `@types/ember`. Since older `ember-source` versions don't expose anything under `ember-source/types`, those side-effect imports will just be silent no-ops in those scenarios.

SergeAstapov:
> those side-effect imports will just be silent no-ops in those scenarios.

if that's true, then it's way to go as it would be simpler

gitKrystan:
This all sounds good. Thanks! I got busy w/ work stuff but will look into it when I'm caught up (if no one else has).

gitKrystan:
@SergeAstapov I think this is ready now.

SergeAstapov:
Thank you @gitKrystan! Looks good to me. Check in app with TS that still on Ember.js 3.28 - all works great.

would like to get approval from @chriskrycho or @dfreeman

dfreeman (approved):
Looks good to me. Thank you for getting a pattern figured out here, @gitKrystan!

SergeAstapov:
CI failure was fixed in #683

PR #683, opened 2023-02-28
Support Ember v5

SergeAstapov:


test-app/config/ember-try.js
```diff
@@ -53,7 +53,10 @@ module.exports = async function () {
         name: 'ember-canary',
         npm: {
           devDependencies: {
-            'ember-source': await getChannelURL('canary'),
+            'ember-source': await getChannelURL('canary').then((url) => {
+              console.log('canary channel url', url);
```
elwayman02, line 57: Did you intend to leave this console.log in?

SergeAstapov: indeed no, that was mistake and was fixed in https://github.com/ember-modifier/ember-modifier/pull/688/files

Issue #613, opened 2022-12-22
3.2+ Cleanup function is not called within specific component


brenner-company:
For a package I'm working on I'm creating a floating-ui modifier & component (using https://floating-ui.com/) to attach floating elements to a trigger/reference. Everything seems to work fine, but I encountered an issue today: as documented within `ember-modifier`, you need to return a cleanup function to remove any added event listeners, etc when the component gets unloaded. 

When I test this with just the modifier, everything works as expected: the cleanup function gets called when eg. navigation away from a page with said modifier. But when I use the component version I've created, the cleanup function is not being called.

I can easily check this by creating the following setup: two toggles that add & remove the modifier and component version. When I open the event listener inspector in Chrome and click the toggle a few times I see the following results.

**Modifier version:**

<img width="422" alt="CleanShot 2022-12-19 at 17 35 04@2x" src="https://user-images.githubusercontent.com/18174827/209193128-42c2a3d9-323b-4d2b-9247-5c9ef5ef2fe1.png">

**Component version:** 

<img width="419" alt="CleanShot 2022-12-19 at 17 36 09@2x" src="https://user-images.githubusercontent.com/18174827/209192048-57f78906-25cc-4f58-a41f-8777b8cc1005.png">

(The event listeners are not getting removed here)

I've put up an example of that code at: https://codesandbox.io/s/cool-morning-9moikc?file=/app/modifiers/au-floating-ui.js (component version can be found under `app/components/au-floating-ui.hbs` & `app/components/au-floating-ui.js`)

By the way: my code is pretty similar (regarding structure) to https://github.com/CrowdStrike/ember-velcro but with less customisation. So, the same issue would also be present in there.

brenner-company:
If, by any chance, the CodeSandbox example preview gets stuck on `Initializing sandbox container`, this can be fixed by toggling (hide + show) the preview pane itself:
![CleanShot 2023-01-02 at 16 42 19@2x](https://user-images.githubusercontent.com/18174827/210258963-27abd87d-1890-496d-a062-d2cafd1204d7.png)

NullVoxPopuli:
:wave: thanks for the reproduction!

any chance the reproduction can be reduced to bare minimum? there is a lot going on in there, and some things looked not wired up correctly :-\

apellerano-pw:
I _think_ I have a bead on this, but it requires Ember 3.27+ for the `(modifier)` helper, so I can't easily produce a twiddle to share.

[...]

PR #725, opened 2023-03-21
Additional tests for #613

NullVoxPopuli:
While debugging https://github.com/ember-modifier/ember-modifier/issues/613#issuecomment-1468823388
I could no longer reproduce the issue.

So... I wonder if there is something specific about that reproduction repo in the comment that is causing problems?
I've kept the tests, because having assurance that certain behaviors work is nice.

### DIFF
--- a/test-app/tests/integration/modifiers/functional-modifier-test.ts
+++ b/test-app/tests/integration/modifiers/functional-modifier-test.ts
@@ -168,6 +168,54 @@ module('Integration | Modifiers | functional modifier', function (hooks) {
       assert.ok(teardownCalls.includes('A'));
       assert.ok(teardownCalls.includes('B'));
     });
+
+    test('teardown is invoked for a dynamically applied modifier', async function (this: Context, assert) {
+      const teardownCalls: string[] = [];
+
+      class State {
+        @tracked isRendered = true;
+        songbird = modifier(() => {
+          return () => teardownCalls.push('destroyed');
+        });
+      }
+
+      const state = (this.state = new State());
+
+      await render(hbs`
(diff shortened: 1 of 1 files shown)


apellerano-pw:
I think the situation only arises if you tear down the rendered DOM while the conditional is still true. The tests here set the conditional back to false, which is why the destructor has a chance to run.

apellerano-pw:
Our team has adopted modifiers pretty quickly, I think because they fill the gap left by Octane when removing `didInsertElement` / `willRemoveElement`. The target for our modifiers is sometimes the `element`, but often not. Take the `on-click-outside` modifier demonstrated in the [Ember Guides](https://guides.emberjs.com/release/components/template-lifecycle-dom-and-modifiers/#toc_out-of-component-modifications) which modifies `document` instead of `element`. It's disastrous if the destructor never runs, amounting to a leak. The more you instantiate the modifier the more click listeners end up weighing down `document`! We use that modifier, and have another similar modifier for adding keyboard shortcuts to the app.

These may in time be considered abuses of modifiers, since they produce side-effects from a certain philosophical perspective. But they're a good way to do DOM lifecycle events in Octane and even encouraged to be used this way by the docs. If Ember were to take a philosophical stance against usages like this I think it would leave a functionality hole in the framework and contradict the docs.

NullVoxPopuli:
I 100% hear you -- and the _need_ for the ability to add listeners to the document is _essential_.
There are def alternate ways to implement the functionality you want, but lemme poke around see if I can confirm this is an oversight with modifier behavior, since it's specific the conditional application of the modifier, I have a hunch: yes -- it's an oversight)

PR #853, opened 2024-03-22
Add TS support for the `modifier` and `modifier-test` blueprints

bertdeblock:
~~Still need to add interfaces to the blueprint files.~~


SergeAstapov:
thank you @bertdeblock! Appears CI is broken, I'll try to fix to be confident in what we release

bertdeblock:
I think I should add https://github.com/ember-polyfills/ember-cli-typescript-blueprint-polyfill as well, otherwise, it won't work for older `ember-cli` versions.

SergeAstapov:
this is published as 4.2.0

PR #854, opened 2024-06-25
Add name to modifier function for Ember Inspector Support

patricklx:
modifiers are now included in the inspector. to be able to display them they need a name. otherwise `<unknown>` will be shown. Currently the source code is displayed instead because of a bug in glimmer...

See also https://github.com/glimmerjs/glimmer-vm/pull/1591

### DIFF
--- a/ember-modifier/src/-private/function-based/modifier.ts
+++ b/ember-modifier/src/-private/function-based/modifier.ts
@@ -134,7 +134,7 @@ export default function modifier(
   // type of `setModifierManager` today is `void`; we pretend it actually
   // returns an opaque `Modifier` type so that we can provide a result from this
   // type which is useful to TS-aware tooling (e.g. Glint).
-  fn.name = name;
+  fn.name = name || fn.name;
   return setModifierManager(
     () => MANAGER,
     fn
--- a/ember-modifier/src/-private/function-based/modifier.ts
+++ b/ember-modifier/src/-private/function-based/modifier.ts
@@ -123,7 +123,7 @@ export default function modifier(
     named: object
   ) => void | Teardown,
   options?: {
-    name: string
+    name: string;
   }
 ): FunctionBasedModifier<{
   Element: Element;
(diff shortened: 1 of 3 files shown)


ember-modifier/src/-private/function-based/modifier.ts
```diff
@@ -121,7 +121,8 @@ export default function modifier(
     element: Element,
     positional: unknown[],
     named: object
-  ) => void | Teardown
+  ) => void | Teardown,
+  name?: string
```
SergeAstapov, line 125: @patricklx as this is a new argument, I guess we should document it in readme to ensure ppl know it exists, how to use and what is the purpose.

@patricklx additionally, v3.2.0+ of this package accepted second argument as an object to aid in v3 -> v4 migrations as described in https://github.com/ember-modifier/ember-modifier/blob/main/MIGRATIONS.md#function-based-modifiers

I'm thinking if we should make function signature like

```
function modifier(
  fn: (
    element: Element,
    positional: unknown[],
    named: object
  ) => void | Teardown,
  options?: { name: string }
```

so that second argument may be expanded again in future if more options can/need to be passed in

ember-modifier/src/-private/function-based/modifier.ts
```diff
@@ -133,6 +134,7 @@ export default function modifier(
   // type of `setModifierManager` today is `void`; we pretend it actually
   // returns an opaque `Modifier` type so that we can provide a result from this
   // type which is useful to TS-aware tooling (e.g. Glint).
+  fn.toString = () => name || fn.name;
```
SergeAstapov, line 137: @patricklx I guess this line should go above the comment `SAFETY: the cast here is ...` to avoid confusion

SergeAstapov (changes requested):

ember-modifier/src/-private/function-based/modifier.ts
patricklx: done

SergeAstapov:
Thank you @patricklx!

SergeAstapov (approved):

SergeAstapov:
@patricklx FYI I'll try to get CI green before publishing this

SergeAstapov:
this is published as 4.2.0

Issue #774, opened 2023-05-15
`eager` option for functional modifiers is undocumented outside migration guide


elwayman02:
The v4 Migration guide mentions that all function-based modifiers must add `{ eager: false }` to migrate to v4. However, there is no mention of function-based modifiers having an options hash in the README for v3 *or* v4, making this an undocumented feature. If functional modifiers are expected to pass this option in v4, it should be documented as part of the core API in the README, not just in the migration guide.

fairnightzz:
I agree with this as well ^

didoo:
+1 on this, expecially in the context of Typescript (I had to add a `@ts-ignore` to prevent TS to complain about this extra argument). Should the signature for the modifiers be updated to include this extra option?

NullVoxPopuli:
>  If functional modifiers are expected to pass this option in v4

they are not.

You remove the option as you upgrade to v4

PR #855, opened 2024-06-28
fix: add `eager` as options to types and docs

aklkv:
related: https://github.com/ember-modifier/ember-modifier/pull/486
fixes: https://github.com/ember-modifier/ember-modifier/issues/774

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,31 +10,32 @@ function-based modifiers and more complicated class-based modifiers.
 
 **NOTE:** this is the README for the v4 release. For the v3 README, see [here](https://github.com/ember-modifier/ember-modifier/blob/v3/README.md).
 
-- [Compatibility](#compatibility)
-  - [TypeScript](#typescript)
-- [Installation](#installation)
-- [Philosophy](#philosophy)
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (changes requested):
The option is not in the current version of ember-modifier -- it was in the previous major.

Only folks using v3 need to add eager false

README.md
```diff
@@ -10,31 +10,32 @@ function-based modifiers and more complicated class-based modifiers.
 
 **NOTE:** this is the README for the v4 release. For the v3 README, see [here](https://github.com/ember-modifier/ember-modifier/blob/v3/README.md).
 
-- [Compatibility](#compatibility)
```
NullVoxPopuli, line 13: Why'd this change?

NullVoxPopuli:
If you wanted to update the v3 readme, that'd be fine 🎉

SergeAstapov:
Agree with @NullVoxPopuli, this option does not exist in v4 hence IMO this is not a legit change

aklkv:
@SergeAstapov yeah we synced up with @NullVoxPopuli on Discord and in alignment. I think I will close this PR and when time permits add notes to v3 docs

PR #959, opened 2025-09-19
Update README.md: Fix "Example with Service Injection"

bantic:
It appears that the example code for "Example with Service Injection" is incorrect: It makes a `this.cleanup()` call, but (unless I'm misunderstanding) there's no `cleanup` method defined on the Modifier class.


NullVoxPopuli (approved):
Tyty

SergeAstapov (approved):

